Issue #159, opened 2015-11-28
Consistency of mef module


castillohair:
I think the following things should be changes from the `mef` module to make it more consistent with itself, and with the terminology used in the paper.

First, looking at the steps involved in standard curve generation:
  - clustering: identification of the beads subpopulations.
  - peak identification: calculation of a statistic from each subpopulation.
  - peak selection: discarding peaks that are too close to saturation.
  - standard curve fitting.

I don't like the following:
- Any reference to "peaks" should be changed to "population" or subpopulation", as in the paper.
- "peak identification" is a terrible name for what is happening here. Identification of the subpopulations takes place in "clustering", and all we do here is calculate a statistic. @JS3xton used a smoothed mode for this matlab code, and I've been using the median for `fc` so far. And some papers recommend using the mean. None of this warrants the name "peak identification". This may even not be a real step anymore.

Additionally, the following changes are proposed:
- in `get_transform_fxn`, arguments `cluster_method`, `find_peaks_method`, and `select_peaks_method` should be allowed to be functions with signatures specified in the docstrings. This is in addition to strings representing the methods included in `mef`. A `standard_curve_model` parameter should be added to fit different types of standard curves (no bead autofluorescence, etc.)
- Clustering method `clustering_dbscan` and `clustering_distance` should be eliminated, as they haven't been tested in a while and they probably underperform.
- `find_peaks_smoothed_mode` should be eliminated. `find_peaks_median` should probably be also eliminated, and standard numpy functions, or functions from the `stats` module should be used.

Issue #146, opened 2015-11-23
get_transform_fxn signature


JS3xton:
From #141, I was confused about the MEF section in `process_beads_table` and some of the confusion seemed to originate from the `get_transform_fxn` signature.

Relevant parts of original thread brought over from #141:

JS:

While I do think it's generally the case that beads should have the same number of peaks, I think it's an unnecessary assumption and possibly detrimental.
- The user may not have put the known MEF values in that way. If this is a required assumption, I think it should be detected and an error should be raised if that assumption is violated. Would this get caught under the current implementation? Or would an underspecified known MEF values just get padded with zeros upon creation of the `mef_values` numpy array?
  - Looks like `mef_values` would _not_ get padded, instead you appear to get a numpy object array where each element is a list. That _may_ still work? I still think the more explicit dictionary data structure would be better there if an error is not warranted.
- It's pretty easy to write this in a manner where each channel is independent. A {channel -> known MEF values} dictionary is intuitive and allows more general specification of known MEF values.
- Some devil's advocate arguments for variable numbers of fluorophore concentrations:
  - it's not outside the realm of possibility that you could get beads with different numbers of fluorophore concentrations for different fluorophores. e.g. 8 MEF but 6 MECY. seems unlikely, though, and I'm not aware of a product which currently exists like this.
  - I could mix two different non-rainbow beads in the same sample, maybe to get FL1 and FL3 coverage, but the beads may have different numbers of peaks.

> SMC:
> - All the MEF values from a single beads sample should have the same number of values, because a single sample has the same number of subpopulations.
> - It seems that you want to change the signature of `get_transform_fxn`, which does not belong here. You might want to open an issue for this. You might also want to reconsider this given my previous point.
> 
> > JS:
> > - Might consider promoting `# Process MEF values` to an emphasized comment.
> >   - I also had to re-read this section several times to understand it. Purpose of `mef_values` and `mef_channels` data structures isn't terribly clear from description or their names. Is `mef_values` referring to _known_ MEF values? It's a list of lists that you convert to a numpy array? That seems dangerous if the specified known MEF values are not all of the same length (different beads have different numbers of peaks. just seems like an unnecessary assumption). Looks like this is supposed to be the input to `mef.get_transform_fxn` which expects a {channel -> known MEF} mapping. I'd almost prefer something more explicit like a dictionary or a list of numpy arrays rather than relying on indexing numpy columns? That would also more generally support different specifications of known MEF values.
[...]

Issue #129, opened 2015-11-17
FCSData interface


castillohair:
Mostly everything that is currently under `channel_info` is awkward to use. We need a better interface to access the data from an FCS file besides the list of events.

PR #165, opened 2015-12-07
MEF module cleanup

castillohair:


fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
+    each group, the 50% samples farther away from the mean are discarded.
+    The mean and covariance are calculated from the remaining samples of
+    each group, and used as initial conditions for the GMM EM algorithm.
 
     `clustering_gmm` internally uses `GMM` from the ``scikit-learn``
-    library. For more information, consult their documentation.
+    library, with full covariance matrices for each cluster and a fixed,
+    uniform set of weights. This means that `clustering_gmm` implicitly
+    assumes that all bead subpopulations have roughly the same number of
+    events. For more information, consult ``scikit-learn``'s documentation.
```
JS3xton, line 70: Throwing away samples 50% away from cluster centers seems pretty arbitrary and unjustified. I assume this is all part of the `scikit-learn` GMM bug? In which case this is kinda a hack to get around that? I've already raised my concerns regarding that bug in #159, and it may not be worth fixing at this point in time. In general, though, it would be nice to get rid of this strange initialization procedure if the `scikit-learn` GMM bug was ever fixed. In my previous experience (MATLAB implementation), a vanilla GMM fit would work fine with a reasonable K-means initialization.

I would also prefer not to hardcode the weights equally. If the GMM algorithm works as advertised, that should be a pretty straightforward parameter to fit, and leaving it unbounded allows for edge cases where you have non-uniform subpopulation sizes (again, the devil's advocate argument that immediately comes to mind is if you mixed two different bead samples and were unable to guarantee equal ratio between the two; just feels like an unnecessary assumption). I feel like this also relates to the GMM bug, though, in which case it may not be worth fixing now and would warrant a deeper investigation into the inner workings of `scikit-learn`'s GMM implementation.

If this works good enough now, then I probably wouldn't change these things.

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
-            covariance_type = 'full', params = 'mc', init_params = '')
-        gmm.weight_ = weights
-        gmm.means_ = means
-        gmm.covars_ = covars
-
-    else:
-        raise ValueError('Initialization method {} not implemented.'\
-            .format(initialization))
+            covars.append(np.cov(data_i.T))
+    means = np.array(means)
```
JS3xton, line 102: this was a little unclear to read through. this is minor though. some possible alternative variable names:
- `sorted_i` -> `sorted_dist_idx`? `dist_idx_sorted`?
- `i` -> `subpop_idx`? `cluster_idx`?
- `il` -> `subpop_low_event_idx`, `ih` -> `subpop_high_event_idx`? or `cluster` in place of `subpop`? or `sample` in place of `event`?
- `sorted_i_i` -> `subpop_frac_idx`? `cluster_frac_idx`?
- `data_i` -> `data_subpop`? `data_cluster`? `data_cluster_frac`?

`discard_frac` also seems kinda like a misnomer? You're discarding 50%, not 25%. Could just do `discard_frac/2` everywhere and have `discard_frac=0.5`. This is really not very important, though.

Not really sure I like any of these suggestions. It was just a little confusing at first (`sorted_i_i` is pretty terrible, though, haha).

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
 
     Returns
     -------
-    peak : float
-        Mode of `data`.
-    hist_smooth : array
-        ``(max_val - min_val + 1)``-long array containing the smoothed
-        histogram.
+    selected_mask : boolean array
+        Set of flags indicating whether a population has been selected.
```
JS3xton: Hesitate to describe it as a set when we're not using a set. boolean array is clearly indicated above, so this isn't a real big issue, but I'd just say "Flags indicating ...", get rid of any reference to a data structure.

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
-    return peak, hist_smooth
-
-def find_peaks_median(data):
-    """
-    Find the median of a data array.
+    # Default thresholds
+    if low is None:
+        low = 0.015*populations[0].domain(0)[0]
+    if high is None:
+        high = 0.985*populations[0].domain(0)[-1]
```
JS3xton: Seems like this would break on a list of arrays where `domain` doesn't exist. Seems like you should query `populations[0]` for a `domain` attribute and raise an error if `low` and `high` are not specified _and_ list is not of `FCSData` objects.

This also assumes it's appropriate to only use the first population's `domain` (from the first channel?). More general solution would not make this assumption; calculate a `low` and `high` for each element in `populations`. In practice, they will probably always be the same, but again feels like an unnecessary assumption which can be pretty easily relaxed (might also be able to solve it recursively by calling `selection_std` on all elements in the array if it detects sequence-type).

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
-    else:
-        ValueError('Number of MEF values and channel peaks does not match.')
-    
-    # Discard unknown (NaN) peaks
-    unknown_mef = np.isnan(sel_peaks_mef)
-    n_unknown_mef = np.sum(unknown_mef)
-    if n_unknown_mef > 0:
-        sel_peaks_ch = sel_peaks_ch[np.invert(unknown_mef)]
-        sel_peaks_mef = sel_peaks_mef[np.invert(unknown_mef)]
+    pop_std[pop_std < min_std] = min_std
```
JS3xton, line 186: I don't understand why minimum STD is set to 1.0. Is this a degenerate case? This seems to make assumptions about the data, e.g. that an STD < 1.0 is too low. In general, I'm not sure why that's justified. If you're operating on a transformed `FCSData` object in RFI space and it's a population on the low end, seems like a standard deviation of < 1.0 might be reasonable there? Actually maybe not. That starts at 10^0? = 1? So you still really shouldn't be at a scale where a STD < 1.0 is reasonable. Either way, would just be nice to state why this is done.

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
 
-    return sel_peaks_ch, sel_peaks_mef
+    # Return populations that don't cross either threshold
+    selected_mask = np.logical_and((pop_mean - n_std_low*pop_std) > low,
+                                   (pop_mean - n_std_high*pop_std) < high)
+    return selected_mask
```
JS3xton, line 191: I'm used to seeing the `&` operator. Seems a little more succinct. Any reason you went with `np.logical_and`?

Also kind of ironic that you returned a mask here instead of just the masked `populations` after so much discussion over masks in the `gate` module. Any reason for this?

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
     ----------
-    data : array
-        Nx1 array to calculate the mode from. `data` is assumed to contain
-        only integers.
-    min_val : int, optional
-        Minimum possible value in `data`.
-    max_val : int, optional
-        Maximum possible value in `data`.
+    populations : list of arrays or FCSData objects
+        Populations to select or discard.
```
JS3xton, line 147: It's also not clear what the dimensionality is for the inputs or the outputs. Input = list of 1D arrays? NxD flow cytometry data arrays (as we've described elsewhere)? `FCSData` objects can have multiple channels; no where is channel specified. Is this selection performed on every channel in an `FCSData` object? equivalently, every column in a multi-dimensional array?

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
-def fit_standard_curve(peaks_ch, peaks_mef):
+def fit_beads_autofluorescence(fl_channel, fl_mef):
     """
-    Fit a standard curve to known fluorescence values of calibration beads.
+    Fit a standard curve using a beads model with autofluorescence.
 
     Parameters
     ----------
-    peaks_ch : array
+    fl_channel : array
```
JS3xton, line 199: Minor thing, but we call it `fl_ch` in the Notes section. Might want to be consistent between the two.

fc/mef.py
```diff
@@ -217,265 +52,152 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
-    peaks_mef : array
+    fl_mef : array
         Fluorescence values of bead populations in MEF units.
 
     Returns
     -------
-    sc : function
+    std_crv : function
         Standard curve that transforms fluorescence from channel units to
         MEF units.
```
JS3xton: Might be nice to indicate the function signature here for the returned `std_crv`. Maybe even just the entire function definition as written in the Notes section?

(both `std_crv` and `beads_model`)

fc/mef.py
```diff
@@ -539,54 +261,67 @@ def sc_fun(p,x):
...
     
-    return (sc, sc_beads, sc_params)
-
-def plot_standard_curve(peaks_ch, 
-                        peaks_mef,
-                        sc_beads,
-                        sc_abs,
-                        xlim = (0., 1023.),
-                        ylim = (1, 1e8)):
+    return (std_crv, beads_model, beads_params)
```
JS3xton, line 292: For `fit_beads_autofluorescence()`, I made some changes after the fork a while back which I think may still be relevant:
- initialized fit with log line (from JS3xton/fc@96a311d1e0b04b6a26cfc104d575f468d28dbc2a):
  
  ``` python
  if kp[0] == 0:
      # If lowest peak is a blank, remove it for the log-y line fit to
      # avoid trying to fit a line to log(0) = -Inf
      p = np.polyfit(fp[1:], np.log(kp[1:]), 1)
  else:
      p = np.polyfit(fp, np.log(kp), 1)
  
  params[0] = p[0]
  params[1] = p[1]
  ```
  
  I think this is generally more robust than what @thoreusc originally coded up which just fits the top two peaks to a line. That method is much more sensitive to top two peak values. This method is like fitting a simpler model first (log line with no fluorescence term which happens to have a closed form solution), and then using that to initialized our slightly more complicated model.
- changed `params[2]` to initialize based on the lowest peak in `fl_channel`. From JS3xton/fc@0dc3babb447931e8a860c75b7776f8752a4f7400:
  
  ``` python
  params[2] = np.exp((params[0] * fp[0]) + params[1])
  ```
  
  This basically just solves the bead model for `fl_mef_auto` using the already calculated slope and intercept in conjunction with the first fit peak ("`fp`"; maybe should be `min(fp)` instead of `fp[0]`, as the intention is really the smallest peak value?). This basically initializes `bead_auto` with the value corresponding to the lowest subpopulation summarizing statistic (median here). I think this is a much better / more intuitive initialization than the arbitrary `100` that @thoreusc initially chose.
- I used the linear algebra `norm` to calculate my error function (JS3xton/fc@0dc3babb447931e8a860c75b7776f8752a4f7400):
  
  ``` python
  res = scipy.optimize.minimize(
      fun=lambda p,x,y: np.linalg.norm(np.log(y + p[2]) - ((p[0] * x) + p[1])),
      x0=params,
      args=(fp,kp))
  ```
  
  I think I've always seen norm used in optimization texts. I believe the current implementation differs from norm because it lacks a final square root. While I don't think that changes the extrema, it might change where on the slope up to / down to the extrema the optimization algorithm stops.
- I see you return a lot of lambda functions. Is this a problem?: http://docs.python-guide.org/en/latest/writing/gotchas/#late-binding-closures  I think I tried to address this in my own version here (JS3xton/fc@d616c36cf2239c09954cd2a3969c85eb0ec06dfc):
  
  ``` python
      # Python binds functions as late a possible, so all of the standards
      # curves would behave the same if we just returned standards_curve
      # (they would all rely upon the final resting params values after the
      # loop terminated). functools.partial addresses this problem by
      # binding some of the function parameters at definition time, not
      # call time.
      sc.append(functools.partial(standards_curve, params=params))
  ```
  
  You kind of have 2 layers of functions here (ones explicitly defined and then lambda functions) which is a little confusing, so I'm not actually sure what happens. Seems like your 2nd layer of lambda functions is equivalent to the partial functions I was using? Does that also solve the problem? I'm not sure.

fc/mef.py
```diff
@@ -539,54 +261,67 @@ def sc_fun(p,x):
...
+
+def plot_standard_curve(fl_channel,
+                        fl_mef,
+                        beads_model,
+                        std_crv,
+                        xlim=(0.,1023.),
+                        ylim=(1.,1e8)):
```
JS3xton, line 299: Not a fan of these hardcoded defaults for `xlim` and `ylim`. Don't like referencing 1023 which is specific to our cytometer bit depth or `1e8` which I have no idea where it comes from. I think these values should ~~either not have defaults (forcing the user to specify them), or~~ default to whatever `matpotlib` decides to show (extremes of `fl_channel` and `fl_mef`? which seems appropriate to me if you have no other additional knowledge about what the limits should be), e.g. `=None` and ignore them.

fc/mef.py
```diff
@@ -626,78 +369,104 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
             Transformation function to convert flow cytometry data from
             channel units to MEF. This function has the same basic
             signature as the general transformation function specified in
             ``fc.transform``.
-        clustering_res : dict
+        clustering : dict
             Results of the clustering step, containing the following
             fields:
             labels : array
                 Labels for each element in `data_beads`.
```
JS3xton: You've introduced the word "element" where we were previously using "event" (i.e. a row in an NxD numpy array of flow cytometry data). In the more general `clustering_gmm` and `selection_std` that might be appropriate as the concept of an event isn't really defined for that context, but by the time you get to `get_transform_fxn`, I think we should be back to "events".

fc/mef.py
```diff
@@ -607,14 +353,11 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
     ----------
     data_beads : FCSData object
```
JS3xton, line 374: Can `data_beads` be a vanilla numpy array? Populating the plot filename from `str(data_beads)` seems like it would be a bad idea for a vanilla numpy array. I'm not even sure it's wise for an `FCSData` object (see discussion in #129 regarding `__repr__` and `__str__`).

fc/mef.py
```diff
@@ -607,14 +353,11 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
         Flow cytometry data, taken from calibration beads.
-    peaks_mef : array
+    mef_values : array
         Known MEF values of the calibration beads' subpopulations, for
         each channel specified in `mef_channels`.
     mef_channels : int, or str, or list of int, or list of str
         Channels for which to generate transformation functions.
```
JS3xton, line 380: Still think `mef_values` and `mef_channels` can be more intuitively combined into a `dict` where key=channel (int or str) and value = 1D numpy array. See #146.

fc/mef.py
```diff
@@ -710,9 +479,9 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
     3. Some subpopulations are then discarded if they are close to either
        the minimum or the maximum channel value. In addition, if the MEF
        value of some subpopulation is unknown (represented as a ``NaN`` in
-       `peaks_mef`), the whole subpopulation is also discarded.
+       `mef_values`), the whole subpopulation is also discarded.
     4. The measured fluorescence of each subpopulation is compared with
-       the known MEF values in `peaks_mef`, and a standard curve function
+       the known MEF values in `mef_values`, and a standard curve function
        is generated using the appropriate MEF model.
```
JS3xton, line 505: In step 2 of Notes, "of each subpopulation" and "for each cluster" is confusingly redundant.

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
     if plot_filename is None:
         plot_filename = str(data_beads)
```
JS3xton, line 527: As mentioned above, this seems like a dangerous misuse of the `__str__` function to populate a file name. Happens to work for current implementation, but I don't think the current `FCSData.__str__()` implementation faithfully performs what `__str__()` functions are expected to do (see discussion in #129 regarding `__repr__` and `__str__`).

I think at least referencing `FCSData.infile` would be better. Even that is ambiguous (file-like or string? maybe both work?). I'd be fine with just raising an error if `plot==True` but `plot_filename` isn't specified unless you can come up with a more reliable way of coming up with a filename from `data_beads`.

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
-            **cluster_params)
+    # mef_channels and mef_values should be iterables.
+    if hasattr(mef_channels, '__iter__'):
+        mef_channel_all = list(mef_channels)
+        mef_values_all = np.array(mef_values).copy()
     else:
-        raise ValueError("Clustering method {} not recognized."
-            .format(cluster_method))
+        mef_channel_all = [mef_channels]
+        mef_values_all = np.array([mef_values])
```
JS3xton: Same as in #146. Seems like dangerous and unnecessary assumption to assume you can make a numpy array from `mef_values` (i.e. that they all have same number of peaks).

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+    # Get number of clusters from number of specified MEF values
+    n_clusters = mef_values_all.shape[1]
+
+    # Run clustering function
+    labels = clustering_func(data_beads[:, clustering_channels],
+                             n_clusters,
+                             **clustering_params)
+
+    # Separate events corresponding to each cluster
     labels_all = np.array(list(set(labels)))
```
JS3xton: is `labels_all` just trying to get all the unique labels? Might consider `unique_labels` instead of `labels_all` for clarity and `np.unique()` instead of invoking `set()`, `list()`, and `np.array()`.

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
-    n_clusters = len(labels_all)
-    data_clustered = [data_beads[labels == i] for i in labels_all]
+    populations = [data_beads[labels == i] for i in labels_all]
+
+    # Sort populations based on distance to the origin
+    cluster_dist = [np.sum((np.mean(di[:,clustering_channels], axis=0))**2)
+                    for di in populations]
```
JS3xton: I don't know what a `di` is. `subpop` instead?

Similarly, we keep going back and forth between "cluster" and "subpop"/"subpopulation". I think we should stick with one or the other (I like "subpop" personally, e.g. `subpop_dist`, `subpop_sorted_ind`).

Also, `ind` versus `idx` for index. We should stick with one or the other. I usually use `idx`. If you've caught them all made them `ind`, then that's fine too. I fear we have both littered throughout, though.

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
     if verbose:
-        print("- STEP 1. CLUSTERING.")
-        print("Number of clusters found: {}".format(n_clusters))
-        # Calculate percentage of each cluster
-        data_count = np.array([di.shape[0] for di in data_clustered])
-        data_perc = data_count*100.0/data_count.sum()
-        print("Percentage of samples in each cluster:")
-        print(data_perc)
+        # Calculate and display percentage of events on each population
+        data_count = np.array([di.shape[0] for di in populations])
```
JS3xton: `di` -> `subpop`

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+    if full_output:
+        stats_values_all = []
+        selected_channel_all = []
+        selected_mef_all = []
+        beads_model_all = []
+        beads_params_all =[]
 
     # Iterate through each mef channel
-    for mef_channel, peaks_mef_channel in zip(mef_channel_all, peaks_mef_all):
+    for mef_channel, mef_values_channel in zip(mef_channel_all, mef_values_all):
```
JS3xton: Still don't like use of these `*_all` variable names (e.g. `mef_channel_all`, `mef_values_all`). There's an implied context which I don't remember as I read through this. I don't remember what the alternative is to the `*_all` variables; what's the difference between `mef_channels` and `mef_channel_all` (not to be confused with `mef_channel` used here as an inner loop temporary variable).

I'm not against overwriting `mef_channels` up in the beginning of `get_transform_fxn` if all the `*_all` variables do is make it iterable.

Might also consider `mef_channel` -> `channel` and `mef_values_channel` -> `values`. Again, this feels like it should be iterating over dictionary key-value pairs (see #146).

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+
+        populations_channel = [pi[:, mef_channel] for pi in populations]
```
JS3xton: `pi` -> `subpop`

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+
+        ###
+        # 2. Calculate statistics in each subpopulation.
+        ###
+
+        # Calculate statistics
+        stats_values = [statistic_func(pi, **statistic_params)
+                        for pi in populations_channel]
```
JS3xton: `pi` -> `subpop`

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+            print("    " + str(stats_values))
+
+        ###
+        # 3. Select populations to be used for fitting
+        ###
+
+        # Select populations based on selection_func
+        if selection_func is not None:
+            selected_mask = selection_func([pi for pi in populations_channel],
+                                           **selection_params)
```
JS3xton: `pi` -> `subpop`

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
-                .format(find_peaks_method))
-        # Sort peaks and clusters
-        ind_sorted = np.argsort(peaks_ch)
-        peaks_sorted = peaks_ch[ind_sorted]
-        data_sorted = [data_clustered[i] for i in ind_sorted]
+            selected_mask = np.ones(n_clusters, dtype=bool)
+
+        # Discard values specified as nan in mef_values_channel
+        selected_mask = np.logical_and(selected_mask,
+                                       ~np.isnan(mef_values_channel))
```
JS3xton, line 659: As mentioned above, not sure why you went with masks here when you fought so hard against them in `gate` module. `selection_func` could just as easily return `populations` already filtered by the mask.

Also used to seeing `&`. As mentioned above, any reason for `np.logical_and` over `&`?

fc/mef.py
```diff
@@ -736,230 +506,223 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
-            # Get colors for peaks
-            colors = [fc.plot.cmap_default(i)\
-                                for i in np.linspace(0, 1, n_clusters)]
+            # Get colors for each population. Colors are taken from the default
+            # colormap in fc.plot, if the population has been selected.
+            # Otherwise, the population is displayed in gray.
+            color_levels = np.linspace(0, 1, n_clusters)
+            colors = [fc.plot.cmap_default(level)
+                      if selected else (0.6, 0.6, 0.6)
+                      for selected, level in zip(selected_mask, color_levels)]
```
JS3xton, line 687: this is super small, but i'd put `if selected else (0.6, 0.6, 0.6)` on the same line as `fc.plot.cmap_default(level)` if it fits, otherwise I'd tab it in one, because those two statements are part of the same clause (the value which goes into the list; it's not a `for` clause and it's not a real `if` clause because it's the weird ternary `if`-`else` thing).

fc/mef.py
```diff
@@ -971,44 +734,41 @@ def get_transform_fxn(data_beads, peaks_mef, mef_channels,
...
+        selection_res['channel'] = selected_channel_all
+        selection_res['mef'] = selected_mef_all
         # Fitting results
         fitting_res = {}
-        fitting_res['sc'] = sc_all
-        fitting_res['sc_beads'] = sc_beads_all
-        fitting_res['sc_params'] = sc_params_all
+        fitting_res['std_crv'] = std_crv_all
+        fitting_res['beads_model'] = beads_model_all
+        fitting_res['beads_params'] = beads_params_all
```
JS3xton: This is minor, but these results dictionaries could probably be made with 1-liners:

``` python
clustering_res = {'labels' : labels}
statistic_res = {'values' : stats_values_all}
```

etc.

fc/mef.py
castillohair, line 70: Both using 50% of the samples for each cluster and forcing the weights to be equal are what's currently making GMM work. This was tested with a bunch of beads files taken by almost everybody in lab.

K-means initialization is actually performed by default (check line 486 in https://github.com/scikit-learn/scikit-learn/blob/c957249/sklearn/mixture/gmm.py). And we saw that this didn't work properly.

Using equal weights is actually less of a hack that you would think. It's a well supported, documented option in scikit's GMM. And the model I'm using with full covariance matrices is arguably less restrictive than the default, which uses equal, diagonal covariance matrices (http://scikit-learn.org/stable/modules/generated/sklearn.mixture.GMM.html).

fc/mef.py
castillohair, line 191: `&` is actually overloaded by `bitwise_and` and not `logical_and`:

``` python
In [1]: a = array([1,3,0,5])

In [2]: b = array([2,0,0,6])

In [3]: a&b
Out[3]: array([0, 0, 0, 4])

In [4]: bitwise_and(a,b)
Out[4]: array([0, 0, 0, 4])

In [5]: logical_and(a,b)
Out[5]: array([ True, False, False,  True], dtype=bool)
```

In this case, the arguments will always be booleans, so it probably doesn't matter.

Regarding masks, note that this function is very different from a gating function. This function is discarding entire populations and not individual events. After calling this function in `get_transform_fxn`, the returned mask is used to discard the corresponding mef values. I was doing it all before in this function, but I thought it was cleaner this way.

fc/mef.py
castillohair, line 147: Good point. The input is supposed to be a list of 1D arrays, and it is called in a loop inside `get_transform_fxn`. I guess it could be extended to work with several dimensions if there is a good reason.

fc/mef.py
castillohair, line 292: I don't think any of the three first points would make any difference. This is a relatively easy and smooth optimization problem. Also, the first peak seen by this function is not autofluorescence if the user decided to discard the first peak.

I don't completely understand the last point. Was this a problem that you actually saw? I don't think this is being a problem currently, otherwise running the example excel file would give the same standard curve three times.

fc/mef.py
castillohair, line 299: Actually, `xlim` is explicitly set when the function is called from `get_transform_fxn`. The default value could be eliminated and there would be no difference. Letting matplotlib figure out the appropriate default xlim is a good idea, though.

For `ylim`, I looked at our typical MEF values and saw that they go from ~1e2 to ~1e6. Adding two decades on each side results in the values that I used. If you let matplotlib figure out the limits, you would end up with different limits when you change the gain, which I don't like very much.

fc/mef.py
castillohair, line 659: See above for answers to both.

fc/mef.py
castillohair, line 102: Addressed in 9a292703e509eb975ba130ba133c22ce0532c2ad

fc/mef.py
castillohair, line 186: Some peaks, especially the highest ones when they were near saturation, tend to aggregate mostly on one bin and give a standard deviation of almost zero. This confused the previous version of the peak selection algorithm.

This might not be necessary with this version, which is cleaner and only uses from 1.5% to 98.5% of the range by default. I'll look into this in another issue.

fc/mef.py
castillohair: Addressed in e504451b8afef1c58a305e4e58b34e07f37018ee

fc/mef.py
castillohair: Addressed in e504451b8afef1c58a305e4e58b34e07f37018ee

fc/mef.py
castillohair, line 186: Added comments in e504451b8afef1c58a305e4e58b34e07f37018ee

fc/mef.py
castillohair, line 147: Addressed in e504451b8afef1c58a305e4e58b34e07f37018ee

fc/mef.py
castillohair: Addressed in 3a1d2737f09c6d2fe84cc817b1a22a1b920ee8cc

fc/mef.py
castillohair, line 199: Addressed in 3a1d2737f09c6d2fe84cc817b1a22a1b920ee8cc

fc/mef.py
castillohair, line 299: Removed the default for `xlim` in 588d65db59beb9bc31fb85677526ab843cf083fd

fc/mef.py
castillohair, line 374: Not for now. The domain is used in a couple of places. It would not be hard to change it, but it will require testing, so this would probably need its own issue.

fc/mef.py
castillohair, line 505: Addressed in 24eb041e68c65c9b0c75be63d9cdd5f720f9f065

fc/mef.py
castillohair: Addressed in 24eb041e68c65c9b0c75be63d9cdd5f720f9f065

fc/mef.py
castillohair: Addressed in a0d781fb00f6fcfc238901cb8283e750b8773549

fc/mef.py
castillohair: Addressed in a0d781fb00f6fcfc238901cb8283e750b8773549

fc/mef.py
castillohair: Addressed in a0d781fb00f6fcfc238901cb8283e750b8773549

fc/mef.py
castillohair: Addressed in a0d781fb00f6fcfc238901cb8283e750b8773549

fc/mef.py
castillohair: Addressed in a0d781fb00f6fcfc238901cb8283e750b8773549

fc/mef.py
castillohair, line 527: The problem with `infile` is that it gives the full path, instead of just the filename.

fc/mef.py
castillohair: Addressed in d627cebf3082816ff1e3a9d6887ede358556acbf

fc/mef.py
castillohair: As for d627cebf3082816ff1e3a9d6887ede358556acbf, all `*_all` variables refer to things that will go in the output. I changed them to `*_out` in 2dfcc8e46baae36fd08a9e4bb212ab58e67cb5b6.

fc/mef.py
castillohair: I don't think there is a situation in which they don't have the same number of peaks. Also, given that clustering can be made with several channels, we definitely need the same number of populations in all of these channels.

fc/mef.py
```diff
@@ -217,265 +52,168 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
+    if low is None:
+        if hasattr(populations[0], 'domain'):
+            low = 0.015*populations[0].domain(0)[0]
+        else:
+            raise TypeError("argument 'low' not specified")
+    if high is None:
+        if hasattr(populations[0], 'domain'):
+            high = 0.985*populations[0].domain(0)[-1]
+        else:
+            raise TypeError("argument 'high' not specified")
```
JS3xton, line 173: I think referencing `populations[0]` will throw an error if `populations` is an empty list, which is never checked for. Not sure if we care about that edge case, though.

fc/mef.py
JS3xton, line 191: Indeed it is. This is a mask array, though, which, as you've stated, should always be of type `bool`.

And yes, I was aware that this wasn't gating `FCSData` objects. Abstractly, though, it's still a mask being applied to an array. It's probably fine either way, just stood out as noticeably different than how we apply masks to arrays in other places in the package.

fc/mef.py
```diff
@@ -217,265 +52,168 @@ def clustering_gmm(data, n_clusters = 8, initialization = 'distance_sub',
...
     Parameters
     ----------
-    data : array
-        Nx1 array to calculate the mode from. `data` is assumed to contain
-        only integers.
-    min_val : int, optional
-        Minimum possible value in `data`.
-    max_val : int, optional
-        Maximum possible value in `data`.
+    populations : list of 1D arrays or FCSData objects
```
JS3xton, line 146: Should probably also say "1-channel FCSData objects"? Or else it should handle numpy arrays and `FCSData` objects with multiple columns, which I don't think it currently does?

fc/mef.py
JS3xton, line 292: > I don't think any of the three first points would make any difference. This is a relatively easy and smooth optimization problem.

It has not been demonstrated to me that these don't make a difference. Especially in light of (1) Prabha's data set where her MEF values change based on the inclusion or exclusion of the 0 peak and (2) the suspicious drop off of fit bead autofluorescence around a gain of 400, and (3) an investigation I did into the different optimization algorithms a while back (see https://wiki.rice.edu/confluence/display/~jts6/JS20150713 and https://github.com/JS3xton/fc/issues/5). I don't know if point 1 and point 2 would fix these things, but I do believe they would make for a more consistent, robust initialization which would be less sensitive to any single peak value. It's possible that the convergence tolerance is low enough that the resulting fit won't change much. I believe these initializations are generally better practice, though, and I am willing to go implement them myself if necessary if there is no good reason why these are worse initializations.

I also don't see a reason not to use norm. Euclidean distance is typically used as an error function for these kinds of optimizations, and anything straying from this should be justified (which the current error function is not). The only argument in favor of the current error function that I see is that it's already done. Again, I'm willing to change it if that is the only objection.

> Also, the first peak seen by this function is not autofluorescence if the user decided to discard the first peak.

I almost mentioned this explicitly. Very true. I still think initializing with a non-zero peak (which should at least result in a `bead_auto` term of comparable units/magnitude to the "true" bead autofluorescence) is much more relevant than arbitrarily choosing 100. I think it would actually be interesting to see if this has any affect on the drop in bead autofluorescence at gains <= 400, as this will force the fitting algorithm to push `bead_auto` down from a non-zero peak, instead of almost certainly pushing it _up_ from a value of 100 (which is an order of magnitude low for FL1).

> I don't completely understand the last point. Was this a problem that you actually saw? I don't think this is being a problem currently, otherwise running the example excel file would give the same standard curve three times.

It was a problem for me a long time ago when I tried to generated multiple standards curves in a loop. I just tried it again, and it doesn't seem to be a problem now (three different functions give rise to 3 different transformed values, as expected):

``` python
>>> import FlowCal as fc
>>> import numpy as np
>>> x = np.array([250, 400, 600, 800, 950])
>>> y1 = np.exp(0.00925*x + 2) - 1000
>>> y2 = np.exp(0.00925*x + 4) - 1000
>>> y3 = np.exp(0.00925*x + 6) - 1000
>>> y1
array([  -925.37317702,   -701.13259903,    900.74273134,  11088.38073022,
        47411.85496472])
>>> y2
array([   -448.57821849,    1208.34799189,   13044.6946715 ,
         88321.72336081,  356717.91218758])
>>> y3
array([    3074.48647773,    15317.60719802,   102777.03682009,
         659003.22476616,  2642197.72074637])
>>> sc1, bm1, p1 = fc.mef.fit_beads_autofluorescence(x, y1)
>>> sc2, bm2, p2 = fc.mef.fit_beads_autofluorescence(x, y2)
>>> sc3, bm3, p3 = fc.mef.fit_beads_autofluorescence(x, y3)
>>> p1
array([  9.68649682e-03,   1.56445560e+00,   1.00000000e+02])
>>> p2
array([  9.30639475e-03,   3.94362557e+00,   1.00000000e+02])
>>> p3
array([  9.24987943e-03,   6.00008724e+00,   1.00021559e+03])
>>> sc1(500)
606.49956529292444
>>> sc2(500)
5414.4303158582998
>>> sc3(500)
41151.964740359654
>>> fxn = []
>>> for y in (y1, y2, y3):
...  sc, bm, p = fc.mef.fit_beads_autofluorescence(x,y)
...  fxn.append(sc)
...
>>> fxn[0](500)
606.49956529292444
>>> fxn[1](500)
5414.4303158582998
>>> fxn[2](500)
41151.964740359654
```

I still don't understand why it isn't a problem (again, I'm assuming the 2 layers of functions somehow uses the lambda functions to bind `beads_param` at definition time, not run time), but it appears to be OK as far as I can tell, so I don't think we need to worry about it.

Another related thing that I forgot to mention in the last comment: based on investigation here: [https://wiki.rice.edu/confluence/display/~jts6/JS20150713 and https://github.com/JS3xton/fc/issues/5], I think we should use Nelder-Mead as the fitting algorithm instead of the default BGFS.

fc/mef.py
JS3xton, line 527: That's a fine reason to not use `infile`, then. Would still not use `__str__`; require caller to specify it if we cannot extract it reliably from `data_beads`.

fc/mef.py
JS3xton: > Also, given that clustering can be made with several channels, we definitely need the same number of populations in all of these channels.

I don't understand this. In general, 2D clustering does not require the same number of clusters in both dimensions.

fc/mef.py
JS3xton: still don't understand `np.array(list(set()))` over `np.unique()`

Issue #175, opened 2016-01-06
Add the channel gains of beads to the output excel file within the beads tab


felixekn:
Quickly identifying the gain of bead files allows the user to double check that the beads were measured at the gains which the samples were measured at. This feature in past versions of FC has been very helpful in diagnosing data irregularities.

castillohair:
Commit a77a8c6da7dc0ef047b13fd5cdefdd318932dba1 implements this. In addition to printing the detector voltage (gain) to the beads sheet, the program now also prints the acquisition time and the number of (gated) events. Running it against the example files gives the correct results as far as I can tell.

@felixekn please test it if you can. @taborlab/flowcal-developers, if there is no proposed change or complain within the next 24 hours, I'll merge into develop.

I think #38 would not be hard to implement after this.

PR #178, opened 2016-01-08
Added excel_ui.add_beads_stats() function.

castillohair:


FlowCal/excel_ui.py
```diff
@@ -562,6 +562,53 @@ def process_samples_table(samples_table,
 
     return samples
 
+def add_beads_stats(beads_table, beads_samples):
```
JS3xton, line 569: I like that this function name is verbose and consistent with other function names (e.g. `process_beads_table`). Might consider updating `add_stats` to `add_samples_stats`.

FlowCal/excel_ui.py
```diff
@@ -808,9 +855,14 @@ def run(verbose=True, plot=True):
         plot=plot,
         plot_dir='plot_samples')
 
-    # Add stats to samples table
+    # Add stats to beads table
     if verbose:
         print ""
+        print("Calculating statistics for beads...")
+    add_beads_stats(beads_table, beads_samples)
```
JS3xton, line 863: i think old print statement above (`print ""`) is not compliant with Python 3.

FlowCal/excel_ui.py
```diff
@@ -562,6 +562,53 @@ def process_samples_table(samples_table,
...
+    beads_table['Number of Events'] = [sample.shape[0]
+                                       for sample in beads_samples]
+    beads_table['Acquisition Time (s)'] = [sample.acquisition_time
+                                           for sample in beads_samples]
+
+    # List of channels that require stats columns
+    r = re.compile(r'^\s*(\S*)\s*MEF\s*Values\s*$')
+    headers = list(beads_table.columns)
+    stats_headers = [h for h in headers if r.match(h)]
+    stats_channels = [r.search(h).group(1) for h in stats_headers]
```
JS3xton: this obscure regex code has been invoked multiple times now (up to 5 by my count?). Might consider pulling it out as a module-level function? maybe hidden?

JS3xton:
Didn't run this, just looked at the source. Minor comments, but otherwise looks good to me if someone else has checked that it produces the right output.

JS3xton:
related to #175

FlowCal/excel_ui.py
castillohair, line 863: Oops. Fixed in a5cd6b6282b4d65007095501ad5a78f6c159477b.

FlowCal/excel_ui.py
castillohair, line 569: Done in e383690f453d761903ab43f55417c12f7b844f18.

FlowCal/excel_ui.py
castillohair: They're two different regexes, actually. Moved them as module-level variables in bd122e8d09088028a2f78f715b7c098d9242454f.

Issue #169, opened 2015-12-17
Point Excel_UI to XLSX sheet from command line?


JS3xton:
Would be kind of nice if I could point the Excel UI to my input XLSX sheet from the command line instead of having to select it from the pop-up prompt every time. I'm envisioning something like:

``` shell
python -m FlowCal.excel_ui <XLSX file>`
```

castillohair:
I noticed that you have used `argparse` for your previous scripts. Do you still think it is good? Have you found anything better since? Do you see any reason for us not to use that for this issue?

Other command line arguments that I'm thinking would be `-v` or `--verbose` to activate/deactivate the verbose mode, and `-p` or `--plot` to produce the plots.

castillohair:
Implemented in 28893f7820950e487d7ef98aba53bebf3b8ccca5, together with the `--verbose` and `--plot` flags. Also the program now accepts a file name, but if not specified it shows the open file dialog.

The minor downside is that the default behavior is now changed: Calling `python -m FlowCal.excel_ui` used to run the program in verbose mode and plot mode, now it doesn't. The new behavior is what I think makes sense, so I don't mind breaking the default behavior. The "Run FlowCal" scripts need to be modified to account for this, though.

castillohair:
62286a6bb58bcec323ba0e88ef8cbac7bad778cb updates the "Run FlowCal" scripts.

castillohair:
Decided to add the ability to specify both the input and the output files in 17fff104e48c04db97feb1ea042d28f9fc45a3cd. `-i` or `inputpath` are used to specify the input excel file, whereas `-o` or `--outputpath` are used to specify the output file. If not specified, the defaults are still to show an open file window, and to use "[input_path]_output" as the name of the output file.

JS3xton:
Sorry for late response. Yeah, I would still go with `argparse`; it was quite useful when I needed more sophisticated command line support (although I haven't had to solve this problem in a while, so I'm not sure if anything better has come along).

Not sure if there are any other flags that might be useful. I can't think of any off the top of my head, but I also haven't thought about it much. It should be pretty straightforward to add more if we need them later.

castillohair:
I'll accept this. If we think about some other flag, we can open another issue.

PR #179, opened 2016-01-08
Cmd args

castillohair:


JS3xton:
related to #169

JS3xton:
Didn't test, but source looks good to me.

PR #188, opened 2016-03-09
Added normed_area and normed_height to hist function.

castillohair:


FlowCal/plot.py
```diff
@@ -229,8 +242,10 @@ def hist1d(data_list,
     if ylabel is not None:
         # Highest priority is user-provided label
         plt.ylabel(ylabel)
-    elif normed:
+    elif normed_area:
         plt.ylabel('Probability')
+    elif normed_height:
+        plt.ylabel('Counts (normalized)')
```
JS3xton, line 250: this implies `normed_area` takes precedence if both are specified as `True`? Actually, you say this up in the description, so that seems fine.

FlowCal/plot.py
```diff
@@ -107,9 +108,13 @@ def hist1d(data_list,
...
     histtype : {'bar', 'barstacked', 'step', 'stepfilled'}, str, optional
         Histogram type. Directly passed to ``plt.hist``.
-    normed : bool, optional
+    normed_area : bool, optional
         Flag indicating whether to normalize the histogram such that the
         area under the curve is equal to one.
+    normed_height : bool, optional
+        Flag indicating whether to normalize the histogram such that the
+        sum of all bins' heights is equal to one. Ignored if `normed_area`
+        is True.
```
JS3xton: Might be nice to explicitly tie these two normalization techniques to their appropriate PDF or PMF representation, but that's pretty minor.

JS3xton:
👍

FlowCal/plot.py
castillohair: Docstring updated in 248aa5609b229eed195e6ac89bda1c3a66a3df28.

PR #189, opened 2016-03-09
Auto exponentiation of channels in RFI

castillohair:


FlowCal/transform.py
```diff
@@ -111,9 +125,68 @@ def exponentiate(data, channels = None):
...
+        else:
+            raise ValueError('max_range should be specified')
+    
+    # If amplification_type is not specified, take it from
+    # data.amplification_type
+    if amplification_type is None:
+        if hasattr(data, 'domain'):
+            amplification_type = data.amplification_type(channels)
+        else:
+            raise ValueError('amplification_type should be specified')
```
JS3xton, line 159: you check for `domain` but then don't use it. looks like you should be checking something else.

castillohair, line 159: Fixed in 4b75616f06639dde0b6fe7034acc68e7e90baf42

PR #193, opened 2016-03-11
Excel UI now reports bead parameters.

castillohair:


castillohair:
I will close this without merging, since #194 implements a better solution.

Issue #184, opened 2016-02-29
Robust handling of bad samples


thoreusc:
Right now, FlowCal does not handle bad sample data very well. During batch-processing, the whole process will crash if a bad sample is processed, and the analysis of all of the preceding samples is lost.

I know that this falls into the realm of "stupid user errors" and is not a problem of FlowCal directly. However, it would be much appreciated if un-processable samples were just skipped, or if the analysis results of that sample were just filled with some null values.

I've attached a bad sample for reference.
[Cells187.zip](https://github.com/taborlab/FlowCal/files/151869/Cells187.zip)

castillohair:
Related to #126.

castillohair:
Solved as part of #194.

Issue #130, opened 2015-11-18
Should throw error if you try to MEF fluorescence data collected through linear amplifier (FSC,SSC amplifier shouldn't matter)


JS3xton:
The standard curve currently implemented in `mef.fit_standard_curve()` is crafted to assume that the x values (assumed to be channel values) are logarithmically related to the y values (known MEF values from the datasheet). This should be true if data was collected through a log amplifier.

If a linear amplifier is detected, an error should be thrown if you try to MEF (at least until an appropriate standard curve can be implemented for a linear relationship between channel values and MEF values). Not sure where this error is appropriate, `excel_ui` at least. Not sure if it should go anywhere else, though, because `mef.fit_standard_curve()` doesn't ask for the amplifier type (nor should it, IMO).

castillohair:
I think this is overriden by #192. Is there any problem if I close this issue?

JS3xton:
Seems like #192 should be updated to not refer to floating point if you want to close this one? This issue deals with linear amplifiers in general (independent of the $DATATYPE). Maybe we get linear amplifiers for free with #192 if FlowCal is modular enough (I haven't looked closely). I think floating point already works, so I really think #192 is a rehashing of this issue. I agree that we should handle linear amplifiers, though, not just throw an error (in that sense, this issue should be update).

castillohair:
I can leave both open, but they will both be closed when the two files that we recently got are able to pass through the Excel UI without issues.

castillohair:
#213 is able to convert floating-point linear data to MEF.

Issue #126, opened 2015-11-15
User-friendly errors in `excel_ui`


castillohair:
TODO: list common error situations. Examples are #98 and #99.

castillohair:
Issues #65, #184, and #130 are examples of errors that we could handle better.

I propose a couple of things here:
1. Have a new column in the samples table in the output file, "Notes" (or any other name), that will contain error messages and warnings originated from processing each sample. If processing is correct,  this field will be empty.
2. Create a new set of exceptions, derived from a new class `FlowCalException`, that will be thrown in response to the errors that we are supposed to handle gracefully at least in the Excel UI.
3. During processing of samples, all the code inside the for loop (line 453 in file excel_ui.py) should be wrapped in a `try` block. Any `FlowCalException` thrown will be caught and dealt by writing to the Notes field. A similar thing could be done for beads.

JS3xton:
These proposals sound like good ideas to me.

castillohair:
Solved as part of #194.

Issue #99, opened 2015-10-26
run_excel.py Crashes on file with low sample counts


BrianLandry:
When trimming is attempted on a file with low sample count it throws and error as it should. run_excel.py should handle that error appropriately instead of crashing.

```
Trimming data...
Traceback (most recent call last):
  File "Code/fc/examples/run_excel.py", line 287, in <module>
    main()
  File "Code/fc/examples/run_excel.py", line 164, in main
    for di in data]
  File "c:\users\brian\onedrive\work\taborlab\code\fc\fc\gate.py", line 41, in start_end
    ' number.')
ValueError: Number of events to discard greater than total number.
```

Example file: (its a .txt so github would allow uploading, change it to fcs after download)
[cells009.txt](https://github.com/castillohair/fc/files/21132/cells009.txt)

PR #194, opened 2016-03-13
Excel ui exceptions

castillohair:
Addresses #184, #126, #99, #98, #65, and #38, and probably some more. It would potentially handle #130, although we now want to support linear amplifiers.

Introduces handling of (some) exceptions without crashing. Examples of things that are now covered are:
- missing file path
- MEF transformation requested in a channel for which a standard curve does not exist
- low number of events
- acquisition settings of beads and samples are different

Instead of crashing, the Excel UI now aborts execution of the current sample and prints the error to both the console (if verbose mode is on) and the Excel output file, column "Successful?".

This has the interesting side effect that some of our example files are no longer valid, since some beads and samples have not been acquired at the same detector voltage.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -254,9 +254,9 @@ def process_beads_table(beads_table,
             except IOError:
                 raise ExcelUIException("file \"{}\" not found".format(
                     beads_row['File Path']))
-            # Check that the number of events is greater than 500
-            if beads_sample.shape[0] < 500:
-                raise ExcelUIException("number of events is lower than 500")
+            # Check that the number of events is greater than 400
+            if beads_sample.shape[0] < 400:
+                raise ExcelUIException("number of events is lower than 400")
 
             ###
             # Transform
@@ -529,9 +529,9 @@ def process_samples_table(samples_table,
             except IOError:
                 raise ExcelUIException("file \"{}\" not found".format(
                     sample_row['File Path']))
-            # Check that the number of events is greater than 500
-            if sample.shape[0] < 500:
-                raise ExcelUIException("number of events is lower than 500")
+            # Check that the number of events is greater than 400
+            if sample.shape[0] < 400:
+                raise ExcelUIException("number of events is lower than 400")
 
             ###
             # Transform
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -223,150 +223,170 @@ def process_beads_table(beads_table,
                         for h in mef_headers_all]
 
     # Iterate through table
+    # We will look for a ExcelUIException on each iteration. If an exception
+    # is caught, it will be stored in beads_samples.
     for beads_id, beads_row in beads_table.iterrows():
+        try:
+            ###
+            # Instrument Data
+            ###
+            # Get the appropriate row in the instrument table
+            instruments_row = instruments_table.loc[beads_row['Instrument ID']]
+            # Scatter channels: Foward Scatter, Side Scatter
+            sc_channels = [instruments_row['Forward Scatter Channel'],
+                           instruments_row['Side Scatter Channel'],
+                           ]
+            # Fluorescence channels is a comma-separated list
+            fl_channels = instruments_row['Fluorescence Channels'].split(',')
+            fl_channels = [s.strip() for s in fl_channels]
 
-        ###
-        # Instrument Data
-        ###
-        # Get the appropriate row in the instrument table
-        instruments_row = instruments_table.loc[beads_row['Instrument ID']]
-        # Scatter channels: Foward Scatter, Side Scatter
-        sc_channels = [instruments_row['Forward Scatter Channel'],
-                       instruments_row['Side Scatter Channel'],
-                       ]
-        # Fluorescence channels is a comma-separated list
-        fl_channels = instruments_row['Fluorescence Channels'].split(',')
-        fl_channels = [s.strip() for s in fl_channels]
-
-        ###
-        # Beads Data
-        ###
-        if verbose:
-            print("\nBeads ID {}...".format(beads_id))
-            print("Loading file \"{}\"...".format(beads_row['File Path']))
-
-        filename = os.path.join(base_dir, beads_row['File Path'])
-        beads_sample = FlowCal.io.FCSData(filename)
-
-        ###
-        # Transform
-        ###
-        if verbose:
-            print("Performing data transformation...")
-        # Transform FSC/SSC to linear scale
-        beads_sample = FlowCal.transform.to_rfi(beads_sample, sc_channels)
-
-        # Parse clustering channels data
-        cluster_channels = beads_row['Clustering Channels'].split(',')
-        cluster_channels = [cc.strip() for cc in cluster_channels]
-
-        ###
-        # Gate
-        ###
-        if verbose:
-            print("Performing gating...")
-        # Remove first and last events. Transients in fluidics can make the
-        # first few and last events slightly different from the rest.
-        beads_sample_gated = FlowCal.gate.start_end(beads_sample,
-                                                    num_start=250,
-                                                    num_end=100)
-        # Remove saturating events in forward/side scatter. The value of a
-        # saturating event is taken automatically from
-        # `beads_sample_gated.domain`.
-        beads_sample_gated = FlowCal.gate.high_low(beads_sample_gated,
-                                                   channels=sc_channels)
-        # Density gating
-        beads_sample_gated, __, gate_contour = FlowCal.gate.density2d(
-            data=beads_sample_gated,
-            channels=sc_channels,
-            gate_fraction=beads_row['Gate Fraction'],
-            full_output=True)
-
-        # Plot forward/side scatter density plot and fluorescence histograms
-        if plot:
+            ###
+            # Beads Data
+            ###
             if verbose:
-                print("Plotting density plot and histogram...")
-            # Define density plot parameters
-            density_params = {}
-            density_params['mode'] = 'scatter'
-            density_params['xlog'] = bool(
-                beads_sample_gated.amplification_type(sc_channels[0])[0])
-            density_params['ylog'] = bool(
-                beads_sample_gated.amplification_type(sc_channels[1])[0])
-            density_params["title"] = "{} ({:.1f}% retained)".format(
-                beads_id,
-                beads_sample_gated.shape[0] * 100. / beads_sample.shape[0])
-            # Plot
-            figname = os.path.join(base_dir,
-                                   plot_dir,
-                                   "density_hist_{}.png".format(beads_id))
-            plt.figure(figsize = (6,4))
-            FlowCal.plot.density_and_hist(
-                beads_sample,
-                beads_sample_gated, 
-                density_channels=sc_channels,
-                hist_channels=cluster_channels,
-                gate_contour=gate_contour, 
-                density_params=density_params,
-                hist_params={'div': 4},
-                savefig=figname)
+                print("\nBeads ID {}...".format(beads_id))
+                print("Loading file \"{}\"...".format(beads_row['File Path']))
 
-        # Save sample
-        beads_samples.append(beads_sample_gated)
+            filename = os.path.join(base_dir, beads_row['File Path'])
+            beads_sample = FlowCal.io.FCSData(filename)
 
-        ###
-        # Process MEF values
-        ###
-        # For each fluorescence channel, check whether a list of known MEF
-        # values of the bead subpopulations is provided in `beads_row`. This
-        # involves checking that a column named "[channel] MEF Values" exists
-        # and is not empty. If so, store the name of the channel in
-        # `mef_channels`, and the specified MEF values in `mef_values`.
-        ###
-        mef_values = []
-        mef_channels = []
-        for fl_channel in fl_channels:
-            if fl_channel in mef_channels_all:
-                # Get header from channel name
-                mef_header = mef_headers_all[mef_channels_all.index(fl_channel)]
-                # Extract text. If empty, ignore.
-                mef_str = beads_row[mef_header]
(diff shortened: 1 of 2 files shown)


FlowCal/excel_ui.py
```diff
@@ -181,16 +188,16 @@ def process_beads_table(beads_table,
...
         A dictionary of MEF transformation functions, indexed by
         ``beads_table.index``.
-    mef_outputs : OrderedDict
-        A dictionary with the intermediate results of the generation of the
-        MEF transformation functions, indexed by ``beads_table.index``.
-        Only included if `full_output` is True.
+    mef_outputs : list
+        A list with intermediate results of the generation of the MEF
+        transformation functions, indexed by ``beads_table.index``. Only
+        included if `full_output` is True.
```
JS3xton, line 194: what was the reason for changing this to a list? I'm not against it, just wondering if there was a motivating reason.

FlowCal/excel_ui.py
```diff
@@ -216,150 +223,181 @@ def process_beads_table(beads_table,
...
+                print("\nBeads ID {}...".format(beads_id))
+                print("Loading file \"{}\"...".format(beads_row['File Path']))
+
+            # Attempt to open file
+            filename = os.path.join(base_dir, beads_row['File Path'])
+            try:
+                beads_sample = FlowCal.io.FCSData(filename)
+            except:
+                raise ExcelUIException("file \"{}\" not found".format(
+                    beads_row['File Path']))
```
JS3xton, line 256: can this `except` be more specific? if it's specific to file not found, is that an IOError? seems slightly more robust to only catch whatever your intended error is and pass other errors through.

FlowCal/excel_ui.py
```diff
@@ -216,150 +223,181 @@ def process_beads_table(beads_table,
...
+            # Check that the number of events is greater than 500
+            if beads_sample.shape[0] < 500:
+                raise ExcelUIException("number of events is lower than 500")
```
JS3xton: where did 500 come from? feels kind of unnecessarily hardcoded, which is usually less robust. `start_end()` looks like it only removes 350 events.

FlowCal/excel_ui.py
```diff
@@ -407,6 +446,15 @@ def process_samples_table(samples_table,
...
         the contents of field "Beads ID".
+    beads_table : DataFrame
+        Table specifying beads samples used to generate
+        `mef_transform_fxns`. This is used to check if a beads sample was
+        taken at the same acquisition settings as a sample to be
+        transformed to MEF. For any beads sample and channel for which a
+        MEF transformation function has been generated, the following
+        fields should be populated: ``<channel> Amp. Type`` and
+        ``<channel> Detector Volt``. If `beads_table` is not specified, no
+        checking will be performed.
```
JS3xton, line 457: agree that function should still work if `beads_table` is not specified. Might add `optional` next to the parameter type to be consistent with other optional parameters.

FlowCal/excel_ui.py
```diff
@@ -450,150 +498,231 @@ def process_samples_table(samples_table,
...
+                print("\nSample ID {}...".format(sample_id))
+                print("Loading file \"{}\"...".format(sample_row['File Path']))
+
+            # Attempt to open file
+            filename = os.path.join(base_dir, sample_row['File Path'])
+            try:
+                sample = FlowCal.io.FCSData(filename)
+            except:
+                raise ExcelUIException("file \"{}\" not found".format(
+                    sample_row['File Path']))
```
JS3xton, line 531: same as above. can this `except` be more specific?

FlowCal/excel_ui.py
```diff
@@ -450,150 +498,231 @@ def process_samples_table(samples_table,
...
+            # Check that the number of events is greater than 500
+            if sample.shape[0] < 500:
+                raise ExcelUIException("number of events is lower than 500")
```
JS3xton: same as above. where'd this hardcoded 500 come from?

JS3xton:
I liked the original, more general "Notes" over "Successful?", since "Notes" accommodates other unforeseen messages we may want to add later, but this is minor. Maybe more specifically "Processing Notes" or "Analysis Notes" or "FlowCal Notes"

> - acquisition settings of beads and samples are different

as discussed in #65, I prefer FlowCal still processing samples where the beads and the sample have different acquisition settings and just adding a warning in the "Notes" / "Successful?" field (I don't have a proposed implementation for this, though. Maybe the "Notes"/"Successful?" column is explicitly backed by a data structure used to accumulate feedback, instead of being passed implicitly via Error messages). This modified behavior might be useful for off-by-1 errors if you really want to use an older beads file after a flow cytometer acquisition card swap, for example. I think this is a low usage usage case, though.

FlowCal/excel_ui.py
```diff
@@ -605,13 +734,32 @@ def add_beads_stats(beads_table, beads_samples):
         FCSData objects from which to calculate statistics.
         ``beads_samples[i]`` should correspond to
         ``beads_table.values()[i]``.
+    mef_outputs : list, optional
+        A list with the intermediate results of the generation of the MEF
+        transformation functions, as given by ``mef.get_transform_fxn()``.
+        If specified, ``mef_outputs[i]`` should correspond to
+        ``beads_table.values()[i]``.
```
JS3xton, line 742: might be nice to specify which columns depend on the `mef_outputs` parameters (i.e. which columns will be excluded if you don't provide `mef_outputs`)

JS3xton:
Can you provide a screenshot of what the output looks like now with some erroneous input files?

Also, as mentioned in #38, I think it would be really helpful to add the fit model parameters to the relevant output plots in `bead_plots`. Rounding to shorten them and tucking them in a corner or in a title would be fine with me. I just often find myself wondering what those values are as I peruse the `bead_plots` figures.

JS3xton:
Regarding the solution for #38, how do you know which element in the comma-separated list corresponds with which model parameters? do we specify the model anywhere? Could a string representation of the model go in the "About Analysis" tab maybe?

FlowCal/excel_ui.py
castillohair, line 194: I changed it to make it more consistent with `beads_samples`. The few times that I've had to use `mef_outputs` (like for some analysis for the paper), I've had to iterate through it along with `beads_samples` and `beads_table`. This is annoying if `mef_outputs` is a dictionary. However, one could do `mef_outputs.items()` and, because it's an `OrderedDict`, the items would be in order. But I haven't had a single case in which I needed the dictionary functionality. Another reason is simplicity. I think it gets confusing if `beads_table` is a dataframe, `beads_samples` is a list, and `mef_outputs` is an ordered dictionary.

FlowCal/excel_ui.py
castillohair, line 256: Fixed in 8640e9fc991c106c4ced517a7d705f5c4c3c7e47. The only error I was contemplating was when the file is not found, which raises an `IOError`.

FlowCal/excel_ui.py
castillohair, line 531: Fixed in 8640e9fc991c106c4ced517a7d705f5c4c3c7e47 as above.

FlowCal/excel_ui.py
castillohair: I figured anything above 350 would be ok. The number is arbitrary, could have been 400, but I think there needs to be some room for the subsequent steps. If you have any suggestions on how to make this more objective, I'm all ears.

FlowCal/excel_ui.py
castillohair: See above.

FlowCal/excel_ui.py
castillohair, line 457: Done in d8e586cbf682958c46072b0f84f36108b59c4e7b.

FlowCal/excel_ui.py
castillohair, line 742: Added in 59e5205f74b65efb8ac9380ff18cb2fa6b2688a7.

castillohair:
Screenshot when running the example files:
image
(Note that some columns are hidden)

Error in S0001 was artificially produced by specifying a nonexisting file. Errors in S0004 and S0005 are actual issues with out crappy example files that we have ignored so far.

castillohair:
> I liked the original, more general "Notes" over "Successful?", since "Notes" accommodates other unforeseen messages we may want to add later, but this is minor. Maybe more specifically "Processing Notes" or "Analysis Notes" or "FlowCal Notes"

Yeah I'm not entirely sold on the name either. "Analysis Notes" sounds good. If I can't come up with anything better in the next few hours, I'll change it to "Analysis Notes".

> > acquisition settings of beads and samples are different
> 
> as discussed in #65, I prefer FlowCal still processing samples where the beads and the sample have different acquisition settings and just adding a warning in the "Notes" / "Successful?" field (I don't have a proposed implementation for this, though. Maybe the "Notes"/"Successful?" column is explicitly backed by a data structure used to accumulate feedback, instead of being passed implicitly via Error messages). This modified behavior might be useful for off-by-1 errors if you really want to use an older beads file after a flow cytometer acquisition card swap, for example. I think this is a low usage usage case, though.

I disagree with this. I think the Excel UI should not allow processing of incorrect beads, because it is guaranteed that this will lead to incorrect MEF values. The edge case you propose is not enough justification. If the user really wants to do this, he can use the python API.

A system for reporting warnings that do not crash processing of a sample is a good idea though. However, I don't think we should go through the trouble of implementing this until we have an actual warning to display.

> Regarding the solution for #38, how do you know which element in the comma-separated list corresponds with which model parameters? do we specify the model anywhere? Could a string representation of the model go in the "About Analysis" tab maybe?

I was thinking about specifying the model in the documentation only. 

If we want to include this in the Excel file, a string representation of the model is needed. I think this should come out of `mef.get_transform_fxn` and should be a function of the chosen fitting function. I'm not sure how to do this. Another option would be to hardcode it in the Excel UI? Overall I'm not sure how to do this. If this is done, I would favor one column with its own model string (since different samples would potentially use different models).

> Also, as mentioned in #38, I think it would be really helpful to add the fit model parameters to the relevant output plots in bead_plots. Rounding to shorten them and tucking them in a corner or in a title would be fine with me. I just often find myself wondering what those values are as I peruse the bead_plots figures.

This requires us to figure out the previous point first.

FlowCal/excel_ui.py
JS3xton: ~~350 is at least justifiable, although there's no direct way to tie it to the default values for `gate.start_end()` (that I know of), so it's effectively still hardcoded. You might also want at least as many non-gated data points as model parameters, but I really don't think anyone is going to have 1 event per subpopulation, so this probably isn't terribly important. If we did 350, would just want to make sure FlowCal doesn't fail catastrophically with only 1 or 2 events (if it did, should figure out why and increase the minimum number of events to mitigate).~~

actually, I'm not even sure why this error check exists. I think we should just defer to any error messages that are raised by any downstream processes that fail (which should ideally give a more detailed explanation for why there were too few events).

JS3xton:
> I disagree with this. I think the Excel UI should not allow processing of incorrect beads, because it is guaranteed that this will lead to incorrect MEF values. The edge case you propose is not enough justification. If the user really wants to do this, he can use the python API.

OK. yeah, you bring up some good points, and  i'm not willing to fight for this, so that's fine by me.

> A system for reporting warnings that do not crash processing of a sample is a good idea though. However, I don't think we should go through the trouble of implementing this until we have an actual warning to display.

agree. wait till we need it.

> I was thinking about specifying the model in the documentation only.
> 
> If we want to include this in the Excel file, a string representation of the model is needed. I think this should come out of `mef.get_transform_fxn` and should be a function of the chosen fitting function. I'm not sure how to do this.

agree with `mef.get_transform_fxn` packaging/maintaining a model string (which maybe should come from the model fitting function (e.g. `fit_beads_autofluorescence()`) in the `mef` module).

> Another option would be to hardcode it in the Excel UI? Overall I'm not sure how to do this. If this is done, I would favor one column with its own model string (since different samples would potentially use different models).

don't like hardcoding it in the Excel UI. this also raises another question: how do you specify different models? That's currently not supported, right? I started doing this in a small Excel-based flow database where I had a "Models" tab where you could specify different models. Not sure if that would be appropriate here. Could be an interesting way to expose an option for specifying your own Model via the Excel UI, though (enter a string, and Excel UI assumes you've provided a Python module with that name which has a model fitting function specified in it).

FlowCal/excel_ui.py
castillohair: My approach has been to just catch the most common errors. The point of this error check in particular is to make sure that the FCS file has enough events for all the gating steps downstream. Having files with low event counts is apparently something people do quite frequently for some reason.

Alternatively, we could wrap the gating function in a `try..except`, and throw an `ExcelUIException` with a more user-friendly message. This could be done for all gating functions. However, I find it easier to just check for a defined number of events at the beginning and say that we don't support files with less events than this number.

> I think we should just defer to any error messages that are raised by any downstream processes that fail (which should ideally give a more detailed explanation for why there were too few events).

So do you think we should have a general `except` statement at the end of the iteration that captures any error that we haven't though of, and prints it to the Excel UI? I'm not against that, but it is not what I was trying to do.

castillohair:
Regarding how get a string representation of the model, I was thinking that the fitting function (`mef.fit_beads_autofluorescence` being the only one we have now) could also output two additional elements. These would be strings that contain the model's equations and the order in which parameters are presented. This could be returned in the `.fitting` field of the output of `get_transform_fxn`, then read by the Excel UI and printed.

castillohair:
As for 48ff173cbcdd0796f66a473e33242496c90121b0, the following changes have been implemented:
- The column "Successful?" has been changed to "Analysis Notes".
- The "Beads" column now reports the beads model equation and the parameter names, using the method I proposed in the last response.
- The lower limit on number of events has been decreased to 400.

I'll merge at the end of the day if there are no more comments.

Issue #192, opened 2016-03-10
Properly process FCS files taken with high-resolution ADCs


castillohair:
We need to make sure that these files can pass through the Excel UI workflow, including MEF, at the very least.

These files are important because new instruments do not have logarithmic amplifiers. Instead, they pass PMT signals through a very high-resolution (18-bit typically) ADC, and store the event signals as floating points. Not supporting these files would reduce our audience significantly.

castillohair:
One big issue here is the `domain` and `hist_bin_edges` properties of `FCSData`. In floating point, the range keyword `$PnR` can be really big. For our new test files, `PnR = 262144`. This can lead to slow performance and memory problems, particularly when plotting 2d density plots or performing density gating. Also, we have seen that some events can be less than zero. And because floating points have (conceptually at least) infinite resolution, `domain` kind of loses its meaning.

One solution is to eliminate `domain` and `hist_bin_edges` when data is in floating point, and give reasonable defaults for number of bins to plotting and density gating functions. I'm open to alternative suggestions.

JS3xton:
Where are `domain` and `hist_bin_edges` used?

Perhaps `hist_bin_edges` can be automatically subsampled in some
intelligent way (that was its original purpose after all).

Can `domain` be a generator? Otherwise it feels like we only use the
cardinality of the domain set; perhaps we just expose $PnR? Can't remember
all the places where `domain` is used.

castillohair:
`domain` is used:
- `high_low` gate, to get the default low/high values.
- mef module, when choosing which peaks to discard based on which events are saturated.
- plot module, to get the `xrange` and `yrange` of several plots.

In all of these cases, only the minimum and the maximum values are used.

`hist_bin_edges` is used:
- density gate
- `hist` and `density2d` in the `plot` module
- `excel_ui`, to produce the histograms tab

castillohair:
I was thinking about discarding both `domain` and `hist_bin_edges`; and incorporating new arguments `range` and `resolution`. `range` would be a 2-element list that would originally go from 0 to `$PnR - 1`, and would be modified with each transformation in a similar way to what `domain` and `hist_bin_edges` do now. `resolution` would be copied from `$PnR` and never modified.

[...]

PR #196, opened 2016-03-21
FCSData histogram bins

castillohair:
Implements and solves #192.

A few things:
- I decided to go with `range` instead of `endpoints` or `span`. It seems to be the most appropriate term for what we want to use, despite the name of the `$PnR` keyword. For example (http://www.dictionary.com/browse/range?):

> range: the extent to which or the limits between which variation is possible.
- `hist_bins` is a function of `FCSData` that generates bins based on `range` and `resolution`, either linearly or logarithmically spaced. One thing that I still don't know how to deal with is when the lower value of `range` is zero and the caller is asking for logarithmically-spaced bins.
- All tests pass. Output has been checked to be identical to the current `develop`.


FlowCal/plot.py
```diff
@@ -590,15 +576,15 @@ def scatter2d(data_list,
     elif hasattr(data_plot, 'channels'):
         plt.ylabel(data_plot.channels[1])
 
-    # Set plot limits if specified, else extract range from domain
+    # Set plot limits if specified, else extract range from range
```
JS3xton: this comment doesn't actually make a lot of sense now that you've swapped "domain" for "range", lol

FlowCal/plot.py
```diff
@@ -379,20 +367,18 @@ def density2d(data,
...
+            and data_plot.hist_bins(1) is not None):
+        # Duplicate bins into a two-element list if necessary
+        if not isinstance(bins, list):
+            bins = [bins, bins]
+        # X Axis
+        if bins[0] is None or isinstance(bins[0], int):
+            bins[0] = data_plot.hist_bins(0, nbins=bins[0], log=xlog)
+        # Y axis
+        if bins[1] is None or isinstance(bins[1], int):
+            bins[1] = data_plot.hist_bins(1, nbins=bins[1], log=ylog)
```
JS3xton: - i don't understand this section. you were originally checking if `bins` is `None` (i.e. only extracting bins from data object if `bins` was not specified). Shouldn't this still be the desired behavior?
- can you add the name of the first argument to `hist_bins()`? (as as you've done with the other 2 arguments) i forgot what that was and had to go look it up.
- was unclear what `data_plot.hist_bins(#)` was doing. still not 100% sure. why are you checking if that's none? when does that function return `None`?
- description of `bins` also didn't make sense. Seems like you should only extract `bins` from data when `bins` is `None`
- does `isinstance(bins,list)` fail with tuples or array-like? should that just be checking for iterable?

FlowCal/plot.py
```diff
@@ -96,16 +95,11 @@ def hist1d(data_list,
...
-        as ``data_list[i].domain``. `div` is ignored if `bins` is
-        specified.
-    bins : array_like, optional
-        bins argument to pass to plt.hist. If not specified, `hist1d`
-        attempts to extract bins from ``data_list[i].domain``.
+    bins : int or array_like, optional
+        If `bins` is an integer, it specifies the number of bins to use.
+        If `bins` is an array, it specifies the bin edges to use. If `bins`
+        is None or an integer, `hist1d` will attempt to use
+        ``data.hist_bins`` to generate the bins automatically.
```
JS3xton, line 108: i don't understand how these two are both true:

> If `bins` is an integer, it specifies the number of bins to use.
> 
> If `bins` is None or **an integer**, `hist1d` will attempt to use `data.hist_bins` to generate the bins automatically.

you have two different sets of instructions for the same condition. may still be true, but it makes description unclear

how is `data.hist_bins` used if `bins` is integer? it uses `hist_bins` as context? defining the range over which to populate the specified number of bins? again, unclear from description.

FlowCal/plot.py
JS3xton: > can you add the name of the first argument to hist_bins()? (as as you've done with the other 2 arguments) i forgot what that was and had to go look it up.

this is actually true all over the place here; anytime you're using a `FCSData.fxn(channel=x)`, would help remind me what that's doing if you prepended the `channel=` part in the function call.

FlowCal/io.py
```diff
@@ -1160,11 +1163,88 @@ def hist_bin_edges(self, channels=None):
...
+        These cover the range specified in ``FCSData.range(channels)`` with
+        a number of bins `nbins`, either linearly or logarithmically
+        spaced.
+
+        Parameters
+        ----------
+        channels : int, str, list of int, list of str
+            Channel(s) for which to get the histogram bins. If None, return
+            a list with bins for all channels, in the order of the
+            `channels` attribute.
```
JS3xton: "attribute" -> "parameter" if i'm gonna be pedantic. I think of "attributes" with objects; parameters as inputs to functions

FlowCal/io.py
```diff
@@ -1160,11 +1163,88 @@ def hist_bin_edges(self, channels=None):
...
+        log : bool, optional
+            Whether to generate bins uniformly spaced in linear or
+            logarithmic scale.
+
+        Return
+        ------
+        array or list of arrays
+            Histogram bins for the specified channel(s).
+
+        """
```
JS3xton, line 1220: - how do you define a bin? bin edge? bin center? or is this quite literally a set of bin objects. unclear.
- typo in short summary

FlowCal/io.py
```diff
@@ -1095,25 +1097,29 @@ def amplifier_gain(self, channels=None):
...
-    def domain(self, channels=None):
+    def range(self, channels=None):
         """
-        Get the domain of the specified channel(s).
+        Get the range of the specified channel(s).
 
-        The domain is inferred from the $PnR parameter, as
-        ``np.arange($PnR)``. The domain should be transformed along with
-        the data when passed through a transformation function.
+        The range is inferred from the $PnR parameter as ``[0, $PnR - 1]``.
```
JS3xton: isn't the range different for floating point? i thought you said elsewhere it was [0, $PnR]

FlowCal/io.py
```diff
@@ -811,10 +811,12 @@ class FCSData(np.ndarray):
...
-    domain(channels=None)
-        Get the domain of the specified channel(s).
-    hist_bin_edges(channels=None)
-        Get histogram bin edges for the specified channel(s).
+    range(channels=None)
+        Get the range of the specified channel(s).
+    resolution(channels=None)
+        Get the resolution of the specified channel(s).
+    hist_bins(channels=None, nbins=None, log=False)
+        Obtain a set of bins for the specified channel(s).
```
JS3xton: "set of bins" is unclear. bin centers? bin edges? how do you define a bin?

FlowCal/gate.py
```diff
@@ -334,11 +352,19 @@ def density2d(data, channels=[0,1],
...
+            and data_ch.hist_bins(1) is not None):
+        # Duplicate bins into a two-element list if necessary
+        if not isinstance(bins, list):
+            bins = [bins, bins]
+        # X Axis
+        if bins[0] is None or isinstance(bins[0], int):
+            bins[0] = data_ch.hist_bins(0, nbins=bins[0], log=xlog)
+        # Y axis
+        if bins[1] is None or isinstance(bins[1], int):
+            bins[1] = data_ch.hist_bins(1, nbins=bins[1], log=ylog)
```
JS3xton: same comments as below for `plot.density2d()`

> - i don't understand this section. you were originally checking if `bins` is `None` (i.e. only extracting bins from data object if `bins` was not specified). Shouldn't this still be the desired behavior?
> - can you add the name of the first argument to `hist_bins()`? (as as you've done with the other 2 arguments) i forgot what that was and had to go look it up.
> - was unclear what `data_plot.hist_bins(#)` was doing. still not 100% sure. why are you checking if that's none? when does that function return `None`?
> - description of `bins` also didn't make sense. Seems like you should only extract `bins` from data when `bins` is `None`
> - does `isinstance(bins,list)` fail with tuples or array-like? should that just be checking for iterable?

FlowCal/gate.py
```diff
@@ -134,14 +134,14 @@ def high_low(data, channels=None, high=None, low=None, full_output=False):
...
+        if hasattr(data_ch, 'range'):
+            high = [np.Inf if di is None else di[1] for di in data_ch.range()]
             high = np.array(high)
         else:
             high = np.Inf
     if low is None:
-        if hasattr(data_ch, 'domain'):
-            low = [-np.Inf if di is None else di[0] for di in data_ch.domain()]
+        if hasattr(data_ch, 'range'):
+            low = [-np.Inf if di is None else di[0] for di in data_ch.range()]
```
JS3xton, line 126: is there a description somewhere that actually specifies that `range[0]` is the low and `range[1]` is the high? kinda obvious, but might be nice to have that explicitly stated somewhere

FlowCal/excel_ui.py
```diff
@@ -773,17 +781,29 @@ def generate_histograms_table(samples_table, samples):
...
-                                'Bin Values ({})'.format(unit)),
-                               columns[0:len(bins)]] = bins
+                                'Bin Centers ({})'.format(unit)),
+                                columns[0:len(bin_centers)]] = bin_centers
                 # Calculate and store histogram counts
-                bin_edges = sample.hist_bin_edges(channel)
                 hist, __ = np.histogram(sample[:,channel], bins=bin_edges)
                 hist_table.loc[(sample_id, channel, 'Counts'),
-                               columns[0:len(bins)]] = hist
+                               columns[0:len(bin_centers)]] = hist
```
JS3xton, line 1114: why do you need bin centers? seems meaningless; `hist()` functions use edge boundaries, not centers. useful for plotting?

JS3xton:
what's the default # of bins now? i see both 256 and 1024 in different places

FlowCal/plot.py
castillohair: Oops. fixed in 618776a212e6469cfd4d959eeb5febc69637f76d.

castillohair:
For the default # of bins, I've tried to maintain the current behavior, which is to use 1024 for density gating and the histograms in the Excel output file, and 256 for plotting. These have worked well, so I decided to keep them.

FlowCal/plot.py
castillohair: I'm trying to replicate the behavior of `bins` here: http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.histogram2d.html

> - i don't understand this section. you were originally checking if `bins` is `None` (i.e. only extracting bins from data object if `bins` was not specified). Shouldn't this still be the desired behavior?

I believe what I'm doing now is at least equivalent, if not better. If `data_plot` does not have `hist_bins`, there is nowhere that we can generate `bins` from.

> - can you add the name of the first argument to `hist_bins()`? (as as you've done with the other 2 arguments) i forgot what that was and had to go look it up.

Will do.

> - was unclear what `data_plot.hist_bins(#)` was doing. still not 100% sure. why are you checking if that's none? when does that function return `None`?

`data_plot.hist_bins` can generate a requested number of bins using the `nbins` argument. If `nbins` is None, however, `data_plot.hist_bins` will use the resolution of the specified channel.

If the user specifies `bins` as an integer, `plot.hist1d` generates the requested number of bins using `data_plot.hist_bins`. Therefore, if `bins` is None or if it's an integer, it should be passed to `data_plot.hist_bins`.

> - description of `bins` also didn't make sense. Seems like you should only extract `bins` from data when `bins` is `None`

See the previous point.

> - does `isinstance(bins,list)` fail with tuples or array-like? should that just be checking for iterable?

No, actually if `bins` is an array I'm assuming that it specifies a set of bin edges.

FlowCal/io.py
castillohair: Actually, this refers to the `channels` attribute of the FCSData class.

FlowCal/io.py
castillohair, line 1220: These are bin edges, as referred to in http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.histogram.html, where the `bins` parameter can contain "bin edges".

Fixed in 9151eb9a0ecda92820f041a943eb84b56ccfe5a2.

FlowCal/io.py
castillohair: Actually, I ran into an issue here. Assume that the range is `[0, $PnR]` and the resolution is `$PnR`. The current implementation of `hist_bins()` would then generate, by default, `$PnR` bins with weird spacing different than one. We could live with these weird bins, or we could make the resolution `$PnR + 1` with floating points, or we could have the range always be `[0, $PnR - 1]`.

I choose the latter. So far we have two independent examples of floating point files, one of which blatantly disregards the standard's description of `$PnR` (i.e. has obviously a higher resolution ADC). In this file, events seem to be in `[0, $PnR]`, although there are negative numbers. Our other, more standard example _does_ have channels with events in `[0, $PnR - 1]`, while using an ADC that clearly has the resolution specified by `PnR`, as I showed you before.

Ultimately, the precise range for floating points may not be that important, given that the standard specifies that the file may contain numbers outside it. I think what I chose is the best option, but I don't think it's better by much.

FlowCal/io.py
castillohair: Fixed in 9151eb9a0ecda92820f041a943eb84b56ccfe5a2.

FlowCal/gate.py
castillohair: Will add `channels=0` to the `hist_bins` call. Read above for responses to your other comments.

FlowCal/gate.py
castillohair, line 126: Yeah, you're right. This should probably be added to `FCSData.range()` though.

FlowCal/excel_ui.py
castillohair, line 1114: Previously we used "Bin values", but as you said before this is poorly defined. I changed the description to "bin centers` because I think this generalizes better to when the data has a resolution of 2e5 but we use a much lower number of bins (what is the "bin value" then?).

Yeah this is for plotting. I'm not against reporting bin edges instead of bin centers, but this should be another issue. We would have to deal with the fact that the list of bin edges will contain one more value than the event counts, and how to present this to the user in a way that is not confusing.

FlowCal/plot.py
castillohair, line 108: If it is an integer, it defines the number of bins to use. If, in addition, `data.hist_bins` is available, a number `bins` of bins will be generated using this function.

Yeah I can see now that it is a little confusing. What do you think of a list format, such as http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.histogram2d.html?

FlowCal/plot.py
castillohair: I think c211b29cf472f4117d3dfab6f464543df660f21e makes this portion of the code more clear.

FlowCal/io.py
castillohair: We could do "in the order of `FCSData.channels`". I'm not completely convinced this would make sense for someone reading the documentation, but it is better than what we have now, I think.

FlowCal/gate.py
castillohair: Check c211b29cf472f4117d3dfab6f464543df660f21e and response above.

FlowCal/gate.py
castillohair, line 126: Added in cee4302047983fcb9cae3edd5fd031368244471f.

FlowCal/io.py
JS3xton: ahhhh wow ok yeah. missed that. I'm in favor of `FCSData.channels`. agree it's better than what we have now, which I clearly mis-interpreted.  XD

FlowCal/io.py
JS3xton: ahhh ok. yeah i agree; i don't think we understand how floating point bins should be generated yet, and I think you've chosen a good default until we can develop a more appropriate model.

looks like you still have a comment which incorrectly describes range for floating point though: https://github.com/taborlab/FlowCal/blob/cee4302047983fcb9cae3edd5fd031368244471f/FlowCal/io.py#L1339

FlowCal/gate.py
JS3xton: we resolved this in person. see summary above.

FlowCal/excel_ui.py
JS3xton, line 1114: ok. this off-by-one # of bin edges relative to # of counts is only relevant for the `Histograms` tab, right? if so, I don't have strong feelings about it; "bin centers" is fine by me

FlowCal/io.py
castillohair: Fixed in 19aba5c59ab06829977097b884265bc5b7a18c73. Also changed for other similar functions.

FlowCal/plot.py
castillohair, line 108: List format added in 89cde65fde3f97ce3367bc92555c1c92f7249f87.

FlowCal/plot.py
castillohair: I think 89cde65fde3f97ce3367bc92555c1c92f7249f87 implements all the changes that we discussed.

JS3xton:
After discussion with @castillohair, agreed on some things:
- Default behavior for `FCSData.hist_bins()` if `nbins=None` is to use `resolution` as number of bins, which we think is good. Can result in _lots_ of bins for high resolution data, but a good ignorant default.
- Using extra context (i.e. max number of bins that "looks good" on a histogram) should be modularly contained to the domain where it is relevant (e.g. the `plot` module). To facilitate this, we agreed to add in defaults into the `plot` module which calls `FCSData.hist_bins()` with "informed" `nbins` values. On this note, I felt that these defaults, which were previously imposed by the `excel_ui` module, should be pulled out of the `excel_ui` module and entirely contained in the `plot` module; `excel_ui` should rely on whatever (hopefully sensible) defaults the `plot` module settles on.
- User specified `bins` should always trump whatever defaults we settle on.

FlowCal/io.py
castillohair: Fixed in ce513630a54154fc7b520f7828ddc665a0ec70e9.

FlowCal/gate.py
```diff
@@ -334,17 +364,35 @@ def density2d(data, channels=[0,1],
...
+            # ``data_ch.hist_bins()``.
+            if not hasattr(bins, '__iter__'):
+                bins = [data_ch.hist_bins(channels=0, nbins=bins, log=xlog),
+                        data_ch.hist_bins(channels=1, nbins=bins, log=ylog)]
 
     # Make 2D histogram
-    if bins is not None:
-        H,xe,ye = np.histogram2d(data_ch[:,0], data_ch[:,1], bins=bins)
-    else:
-        H,xe,ye = np.histogram2d(data_ch[:,0], data_ch[:,1])
```
JS3xton, line 338: this `np.histogram2d` call will now fail of `bins` is `None`. Can happen if you pass in a numpy array (instead of an `FCSData` object) and don't specify `bins`.

spoke with @castillohair in person, but not sure if we want to update `bins` default for the `plot` and `gate` density functions or add the `if-else` statement back in.

FlowCal/plot.py
```diff
@@ -378,31 +381,44 @@ def density2d(data,
...
 
     # If colormap is not specified, use the default of this module
     if 'cmap' not in kwargs:
         kwargs['cmap'] = cmap_default
 
     # Calculate histogram
-    if bins is not None:
-        H,xe,ye = np.histogram2d(data_plot[:,0], data_plot[:,1], bins=bins)
-    else:
-        H,xe,ye = np.histogram2d(data_plot[:,0], data_plot[:,1])
```
JS3xton, line 411: same as above. this call to `np.histogram2d` will fail if `bins` is `None`.

FlowCal/gate.py
```diff
@@ -334,17 +364,35 @@ def density2d(data, channels=[0,1],
...
         raise ValueError('data should have more than one event')
 
-    # Extract default bins if necessary
-    if (bins is None and hasattr(data_ch, 'hist_bin_edges')
-            and data_ch.hist_bin_edges(0) is not None
-            and data_ch.hist_bin_edges(1) is not None):
-        bins = np.array(data_ch.hist_bin_edges())
+    # If ``data_ch.hist_bins()`` exists, obtain bin edges from it if
+    # necessary.
+    if hasattr(data_ch, 'hist_bins'):
```
JS3xton: `hist_bins` is a function. was thinking it would be better to check if it's callable. that might get kinda kluge, though, if you first need to check if the data object has `hist_bins` and then to check if `hist_bins` is callable. Originally thought there was a 1-test way to check if something is callable, but couldn't find it quickly.

FlowCal/plot.py
```diff
@@ -378,31 +381,44 @@ def density2d(data,
...
-        xdx = np.linspace(0, 1, len(bdx))
-        xsx = np.linspace(0, 1, (len(bdx) - 1)/div + 1)
-        xdy = np.linspace(0, 1, len(bdy))
-        xsy = np.linspace(0, 1, (len(bdy) - 1)/div + 1)
-        # Generate sub-sampled bins
-        bins = np.array([np.interp(xsx, xdx, bdx), 
-                            np.interp(xsy, xdy, bdy)])
+    # If ``data_plot.hist_bins()`` exists, obtain bin edges from it if
+    # necessary.
+    if hasattr(data_plot, 'hist_bins'):
```
JS3xton: same as above. not sure if you want to update this test to acknowledge that `hist_bins` should be callable.

FlowCal/gate.py
castillohair: Fixed in b2b5d3605a13a621070c2ece6e8e9688e6cb7bf9.

FlowCal/plot.py
castillohair: Fixed in b2b5d3605a13a621070c2ece6e8e9688e6cb7bf9.

FlowCal/gate.py
castillohair, line 338: Changed default `bins` from `None` to an appropriate number on `plot.density2d` and `plot.hist1d`, and `gate.density2d` in f304166b28904bed3c8c9ac5f7b88a0bcf7cf225. The defaults were chosen to match what we currently know works (the behavior with the example files would be maintained). In addition, I removed explicit setting of bin values from the Excel UI's code.

FlowCal/plot.py
castillohair, line 411: See above.

castillohair:
All comments addressed. Branch was merged into develop.

PR #200, opened 2016-03-26
to_rfi() has been modified to use linear gain.

castillohair:
Unit tested. Seems to process our only sample with an existing, different than one linear gain well (the time channel in `test/Data004.fcs`).


FlowCal/transform.py
```diff
@@ -83,21 +83,34 @@ def transform(data, channels, transform_fxn, def_channels = None):
...
+def to_rfi(data,
+           channels=None,
+           amplification_type=None,
+           amplifier_gain=None,
+           max_range=None):
     """
     Transform flow cytometry data to Relative Fluorescence Units (RFI).
 
-    This function applies the following transformation:
+    If ``a[0]`` is different from zero, data has been taken using a log
```
JS3xton: unclear what `a[0]` is referring to. I would think this should refer to the parameters listed above in the function signature.

JS3xton: true all throughout this docstring actually. I would expect to just use the parameter name everywhere, but those names are kinda long / kludge. maybe shorten to `amp_type` and `amp_gain`? this is a minor critique, though.

FlowCal/transform.py
```diff
@@ -118,6 +128,15 @@ def to_rfi(data, channels=None, max_range=None, amplification_type=None):
         the first element is zero, the amplification type is linear. If
         None, take `amplification_type` from
         ``data.amplification_type(channel)``.
```
JS3xton: this description of `amplification_type` sounds pretty non-intuitive. might be nice to reference the FCS standards to indicate that it's actually just mimicking the `$PnE` keyword. this is minor, though.

FlowCal/transform.py
```diff
@@ -129,53 +148,98 @@ def to_rfi(data, channels=None, max_range=None, amplification_type=None):
...
+        # Define transformation, depending on at[0]
         if at[0]==0:
-            continue
-        # Define transformation
-        tf = lambda x: at[1] * 10**(at[0]/float(r) * x)
+            # Linear amplifier
+            # If no amplifier gain has been specified, try to obtain from data,
+            # otherwise assume one
+            if ag is None:
+                if hasattr(data, 'amplifier_gain'):
```
JS3xton: again, might wanna check if `amplifier_gain` is callable, but this is minor

FlowCal/transform.py
```diff
@@ -129,53 +148,98 @@ def to_rfi(data, channels=None, max_range=None, amplification_type=None):
...
+                    ag = data.amplifier_gain(channel)
+                    # If the linear gain has not been specified, it should be
+                    # assumed to be one.
+                    if ag is None:
+                        ag = 1.
+                else:
+                    ag = 1.
```
JS3xton, line 226: this whole `if-else` segment seems like it could possibly be more elegantly done with a `try-except`; just try calling `data.amplifier_gain`, and `except: ag=1`. this is minor, though

FlowCal/transform.py
```diff
@@ -129,53 +148,98 @@ def to_rfi(data, channels=None, max_range=None, amplification_type=None):
...
+            tf = lambda x: x/ag
+        else:
+            # Log amplifier
+            # If no range has been specified, try to obtain from data.
+            if r is None:
+                if hasattr(data, 'domain'):
+                    r = len(data.domain(channel))
```
JS3xton: `domain` no longer exists right? i think you need to update this.

FlowCal/transform.py
```diff
@@ -118,6 +128,15 @@ def to_rfi(data, channels=None, max_range=None, amplification_type=None):
...
         ``data.amplification_type(channel)``.
+    amplifier_gain : float or list of floats, optional
+        The linear amplifier gain of the specified channel(s). Only used if
+        ``amplification_type[0]==0`` (linear amplifier). If None,
+        take `amplifier_gain` from ``data.amplifier_gain(channel)``. If
+        `data` does not contain ``amplifier_gain()``, use 1.0.
+    max_range : int, float, or list of int or float, optional
+        Maximum range, for each specified channel. Only needed if
+        ``amplification_type[0]!=0`` (log amplifier). If None, take
+        `max_range` from ``len(data.domain(channel))``.
```
JS3xton: do you call it `max_range` because `range` is already a native Python keyword? `max_range` wasn't intuitive for me, was trying to figure out why you had to specify a max (like why was it bounded? and how did that factor into the `to_rfi()` transformation). I guess i'd prefer `amp_range` (consistent with the shortening of the other 2 amplifier parameters)

FlowCal/transform.py
JS3xton: oh i guess this is an orthogonal change (those changes are being introduced in another branch which hasn't been merged yet). when you merge, though, you'll need to update `domain` and `hist_bin_edges`, but that's a merge issue which should be treated separately

JS3xton:
aright looked through code. did not look through tests, tho

FlowCal/transform.py
castillohair: If you have suggestions to improve readability, I'm all ears. It reads ok to me.

Referencing the `$PnE` sounds more like something that should go in a `Notes` section instead of the description of the argument. I could do this.

FlowCal/transform.py
castillohair, line 226: We've been explicitly checking for things using `hasattr()` throughout the package. For consistency, I would not want to change this to `try`/`except`.

FlowCal/transform.py
castillohair: This is not part of the modifications introduced in this merge request. It is highlighted with green because I moved it down.

FlowCal/transform.py
castillohair: Restricted short names to the paragraphs after equations in cb118e5ee84243eb00320cd431aaa5b34578722b. Everything else now uses the proper names of the arguments.

FlowCal/transform.py
castillohair: Changed in 25bc5b81206d87e8fb61863da1f081ce7bfb7ca9.

FlowCal/transform.py
castillohair: On second thought, it may be worth changing this to "resolution", since that what `r` in the equation is supposed to be. Also, on line 232 and 233, we are taking `max_range` from `data.resolution` if it is not available. Changing this to `resolution` would make it consistent with the rest of our nomenclature, and preserve the nice pattern that this function has in which the argument names correspond to the property names of `FCSData` that they're taken from when they're not specified.

tl;dr: `max_range` should be change to `resolution`.

castillohair: Done in 02681e78dd51d1f0b6fd0ea2e0d7770a8ffda1e1.

FlowCal/transform.py
castillohair: Changed my mind and added the `$PnE` reference as indicated in 818a29410dd35b40a66aecbfd9c0f8682b93fe8d.

castillohair:
I addressed every comment. I'll merge in an hour if there's nothing else.

Issue #174, opened 2016-01-06
Excel UI crashes on OSX with Anaconda 2.4.1


castillohair:
When running FlowCal's Excel UI on OSX with Anaconda 2.4.1, the following error message is shown:

```
2016-01-04 15:19:07.322 python[83254:10454178] -[NSApplication _setup:]: unrecognized selector sent to instance 0x101ca0970
2016-01-04 15:19:07.324 python[83254:10454178] An uncaught exception was raised
2016-01-04 15:19:07.324 python[83254:10454178] -[NSApplication _setup:]: unrecognized selector sent to instance 0x101ca0970
2016-01-04 15:19:07.324 python[83254:10454178] (
    0   CoreFoundation                      0x00007fff931b0ae2 __exceptionPreprocess + 178
    1   libobjc.A.dylib                     0x00007fff9c76873c objc_exception_throw + 48
    2   CoreFoundation                      0x00007fff931b3b9d -[NSObject(NSObject) doesNotRecognizeSelector:] + 205
    3   CoreFoundation                      0x00007fff930ec601 ___forwarding___ + 1009
    4   CoreFoundation                      0x00007fff930ec188 _CF_forwarding_prep_0 + 120
    5   libtk8.5.dylib                      0x0000000104acc76d TkpInit + 472
    6   libtk8.5.dylib                      0x0000000104a44f91 Initialize + 1541
    7   _tkinter.so                         0x0000000104910a3d Tcl_AppInit + 77
    8   _tkinter.so                         0x000000010490e485 Tkinter_Create + 997
    9   libpython2.7.dylib                  0x00000001000c74bc PyEval_EvalFrameEx + 40124
    10  libpython2.7.dylib                  0x00000001000c83a3 PyEval_EvalCodeEx + 2131
    11  libpython2.7.dylib                  0x000000010003e100 function_call + 176
    12  libpython2.7.dylib                  0x000000010000c552 PyObject_Call + 98
    13  libpython2.7.dylib                  0x000000010001dbed instancemethod_call + 333
    14  libpython2.7.dylib                  0x000000010000c552 PyObject_Call + 98
    15  libpython2.7.dylib                  0x00000001000bc547 PyEval_CallObjectWithKeywords + 87
    16  libpython2.7.dylib                  0x000000010002211e PyInstance_New + 126
    17  libpython2.7.dylib                  0x000000010000c552 PyObject_Call + 98
[...]

PR #202, opened 2016-03-26
Excel UI's output file resizes columns automatically.

### LINKED ISSUE #201: Autoadjust widths of Excel output file
A fixed column width in the output file is a little ugly, especially now that the beads sheet is populated with a lot of things.

Unfortunately, automatic column width is not something directly enabled by pandas: https://github.com/pydata/pandas/issues/4049 http://stackoverflow.com/questions/17326973/is-there-a-way-to-auto-adjust-excel-column-widths-with-pandas-excelwriter. However, I think even a simple naive solution based on the length of the strings in each column would be an improvement over the current situation.

castillohair:
Fixes #201.


JS3xton:
I'm not familiar with `tkFont`. Does it come natively? or is that an additional 3rd party package requirement? otherwise this pull request looks fine to me. would be nice to see a before / after, but I trust it. also, did you ever test it on OSX? or Ubuntu/Mint?

castillohair:
It turns out that this does not work perfectly on OSX or Linux. Also, making it dependent on `tk` gives rise to issues in OSX, similar to #174. I decided to revert to the previous version that relied on string length alone. It is not perfect on any platform, but it is a big improvement over what we had before.

Before (Linux Mint):
before

After (Linux Mint):
after

Output in windows is similar.

castillohair:
Comparison on Windows:

Before:
image

After:
image

PR #206, opened 2016-05-09
Improved ability to handle FCS files from different sources.

castillohair:
Solves #203.


FlowCal/io.py
```diff
@@ -164,6 +164,12 @@ def read_fcs_text_segment(buf, begin, end, delim=None):
     buf.seek(begin)
     raw = buf.read((end+1)-begin)
 
+    # Some files have been found to include space characters at the end of the
+    # TEXT segment. If the delimiter is not a space character, strip spaces
+    # from the end of the segment.
+    if not delim.isspace():
+        raw = raw.rstrip()
+
```
JS3xton, line 170: maybe you should cut off everything past the last instance of the delimiter? this is a minor distinction, but cutting off all white space may not be the same as eliminating everything after the last delimiter. if, on the other hand, some files do not include a final delimiter at the end of the TEXT segment, this would throw away valid values, which would be bad.

FlowCal/io.py
```diff
@@ -174,15 +180,40 @@ def read_fcs_text_segment(buf, begin, end, delim=None):
...
+                    s += pairs_list[pairs_list_idx]
+                    break
+            # Append temporary string to previous element if pairs_list_delim
+            # is not empty, otherwise make it the first element.
+            if pairs_list_delim:
+                pairs_list_delim[-1] += s
+            else:
+                pairs_list_delim.append(s)
+        pairs_list_idx += 1
+    pairs_list = pairs_list_delim
```
JS3xton, line 217: i think this part works. couldn't think of any edge cases that might break it.

FlowCal/io.py
```diff
@@ -627,27 +673,35 @@ def __init__(self, infile):
...
-                    delim=delim)[0]
+                try:
+                    self._analysis = read_fcs_text_segment(
+                        buf=f,
+                        begin=analysis_begin,
+                        end=analysis_end,
+                        delim=delim)[0]
+                except:
+                    warnings.warn("ignoring ANALYSIS segment")
+                    self._analysis = {}
```
JS3xton, line 699: would be nice if these warnings could somehow forward the error message. currently doesn't give a very informative warning.

FlowCal/io.py
```diff
@@ -627,27 +673,35 @@ def __init__(self, infile):
...
             else:
                 self._analysis = {}
         else:
             self._analysis = {}
         
         # Import DATA segment
-        param_ranges = [int(self._text['$P{0}R'.format(p)])
+        param_ranges = [float(self._text['$P{0}R'.format(p)])
```
JS3xton, line 706: when is `param_ranges` a float? originally we thought it corresponded to the bit depth of the ADC (which should still be an `int`). Curious to know when it's not an integer. I think `read_fcs_data_segment()` is expecting this to be number of bits (especially if $DATATYPE is 'I').

castillohair, line 706: File `MVa2011-06-30_fcs31.fcs` in our test repo does this (see #203). The standard does not explicitly require `$PnR` to be an integer when the datatype is floating point.

FlowCal/io.py
castillohair, line 170: I like this idea more than what I had. I changed it in 5185f7a692852d5bb6ef74ca44095fb30483a420. The same number of files pass/fail.

FlowCal/io.py
castillohair, line 699: Done in 549b22af821c73e2e3a2709c9e7b08e9b33735ab.

castillohair:
I'll merge by the end of the day if there are no further comments.

FlowCal/io.py
JS3xton, line 706: Ahh ok. It appears that only the `Time` channel reports a floating point `$PnR`, which makes a little more sense.

`$PnR` values constitute an optional parameter to `fc.io.read_fcs_data_segment()` which are only used if `$DATATYPE` is `I` to mask off unused bits (which I've never actually seen change the data, but strictly adheres to the standard). The `$DATATYPE` of this file is `F` (single precision floating point), so the `$PnR` values are never used by `fc.io.read_fcs_data_segment()`. If `$DATATYPE` were `I`, then we'd probably hit an error (how would you even interpret a floating point `$PnR` value, though? This is what I was originally getting hung up on).

Furthermore, I agree with @castillohair that the standard doesn't expressly forbid floating point values for `$PnR` parameters, so the most robust approach to accommodating this wrinkle is to report `param_ranges` as a `float`. Assuming sensible `$PnR` values are still reported when `$DATATYPE` is `I`, they should still serve their purpose (masking off unused bits) as `float`s.

FlowCal/io.py
```diff
@@ -164,21 +164,18 @@ def read_fcs_text_segment(buf, begin, end, delim=None):
...
 
-    # Some files have been found to include space characters at the end of the
-    # TEXT segment. If the delimiter is not a space character, strip spaces
-    # from the end of the segment.
-    if not delim.isspace():
-        raw = raw.rstrip()
+    # Check that the first character of the TEXT segment is equal to the
+    # delimiter.
+    if raw[0] != delim:
+        raise ValueError("segment should start with delimiter")
```
JS3xton, line 174: maybe add a check that `len(raw)>0`? how does this parse an empty segment (e.g. ANALYSIS)? I think it never gets called? cause the `FCSFile` object checks the HEADER and finds offset values of 0, so it never calls `fc.io.read_fcs_text_segment()` with an empty segment? this is kinda an edge case, but would be nice if `read_fcs_text_segment()` succeeded when passed a degenerate segment (e.g. `raw=''` or `raw='//'`, resulting in empty dictionaries?) or failed gracefully when passed a degenerate malformed segment (`raw='/'`).

FlowCal/io.py
```diff
@@ -164,21 +164,18 @@ def read_fcs_text_segment(buf, begin, end, delim=None):
...
 
-    pairs_list = raw.split(delim)
+    # Look for the last delimiter in the segment string, and retain everything
+    # from one character after the first delimiter to one character before the
+    # last delimiter.
+    end_index = raw.rfind(delim)
+    raw = raw[1: end_index]
```
JS3xton, line 180: The edges cases I can think of for this appear to work in an intuitive manner (assuming `delim='/'`):
1. `raw=''` => `end_index=-1` => new `raw=''` => `pairs_list=['']`
2. `raw='/'` => `end_index=0` => new `raw=''` => `pairs_list=['']`
3. `raw='//'` => `end_index=1` => new `raw=''` => `pairs_list=['']`
4. `raw='///'` => `end_index=2` => new `raw='/'` => `pairs_list=['','']`

I think 1-3 will fail later due to:

``` python
# List length should be even since all key-value entries should be pairs
if len(pairs_list) % 2 != 0:
    raise ValueError("odd # of (keys + values); unpaired key or value")
```

Could possibly throw an explicit error message for these, but not sure they warrant their own message.

castillohair, line 180: I'm ok with these edge nonstandard cases failing with weird, slightly non-appropriate error messages.

FlowCal/io.py
castillohair, line 174: Added handling of empty raw string in 78540eddb5496388fe9fa57e5ceebe5f0458b8fc.

Issue #207, opened 2016-05-09
Change 'log' arguments to 'scale'


castillohair:
Some plotting functions have a boolean `log` argument or some variation. Not surprisingly, this argument allows one to specify axes in log scale.

I propose to change all of these to a string `scale`, which could be `linear` or `log`. This allows us to generalize and include more scales (e.g. logicle) if/when we implement them. Furthermore, this aligns better with matplotlib's practices (see arguments `xscale` and `yscale` in http://matplotlib.org/api/axes_api.html, `plt.xscale()` and `plt.yscale()` in http://matplotlib.org/examples/pylab_examples/symlog_demo.html, among others).

Same applies to density gating and `FCSData.hist_bins`.

PR #208, opened 2016-05-09
`log` arguments changed to `scale`

castillohair:
Solves #207.


JS3xton:
Don't see an immediate driving need for this, but I do think it's clearer and more consistent with existing libraries.

Would also be nice if `lin` served as a valid scale string (treated same as `linear`), but this is minor.

castillohair:
matplotlib functions, from which I'm taking this idea of using a `scale` argument, only use `linear` to represent linear scale. I'd rather be consistent with this.

castillohair:
If there are no more comments, I'll merge tonight.

PR #210, opened 2016-05-11
Log scale for scatter plots

castillohair:
Solves #209.

### DIFF
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -49,7 +49,6 @@ Functions in this module are divided in two categories:
 import numpy as np
 import scipy.ndimage.filters
 import matplotlib
-import matplotlib.ticker
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


castillohair:
This is from the file that we have been using to test floating-point linear MEFing:

``` python
FlowCal.plot.scatter2d(populations,
                       channels=fl_channels[:2],
                       xscale='log',
                       yscale='log')
```

image

``` python
FlowCal.plot.scatter3d(populations,
                       channels=fl_channels,
                       xscale='log',
                       yscale='log',
                       zscale='log')
```

image

```
FlowCal.plot.scatter3d_and_projections(
    populations,
    channels=fl_channels,
    xscale='log',
    yscale='log',
    zscale='log')
pyplot.tight_layout()
```

image

With linear scale, the plots are as usual (run the Excel UI example to see).

Issue #195, opened 2016-03-16
Suport MEF transformation with linear amplifier


castillohair:

castillohair:
It turns out that the SEA's flow cytometer can export data in FCS 2.0 in both linear and log scale, and 3.0 in linear scale with floating points. I got all three versions after one of my experiments, including several cell samples and beads. @JS3xton if you're interested in looking at them, I can put them on Dropbox or something.

This issue will be considered resolved when the three datasets result in the same MEF values for each cell sample when run through the Excel UI. There may be some issues in processing integer linear data, so at the very least integer log data and floating-point linear data should work. 

I will now show the result of running the Excel UI with the three sets of files, as performed by the current `floating-mef` branch (43a9dcb5dad54e789e89416e768e838aab01c10e).

# FCS2.0 with log amplifier, 10-bit integer

This is the setup that we currently support, so the following results are what we would consider correct.

## Beads

Initial density/histogram plot:
![density_hist_b0001](https://cloud.githubusercontent.com/assets/1529481/15235240/2d8431f2-187f-11e6-92ee-09cbc6105ece.png)
2D Clustering:
![clustering_b0001](https://cloud.githubusercontent.com/assets/1529481/15235275/74d103e6-187f-11e6-9ef3-30c3053487be.png)
Populations used for standard curve generation:
![populations_apc-a_b0001](https://cloud.githubusercontent.com/assets/1529481/15235241/323b2840-187f-11e6-8de9-77563a2e8f8b.png)
Standard curve:
![std_crv_apc-a_b0001](https://cloud.githubusercontent.com/assets/1529481/15235243/3633f15c-187f-11e6-9950-b98be889bed2.png)

## Cell sample

![s0024](https://cloud.githubusercontent.com/assets/1529481/15235269/5e146ed6-187f-11e6-80a8-847d0ac0befb.png)

# FCS2.0 with linear amplifier, 10-bit integer

## Beads

Initial density/histogram plot:
![density_hist_b0002](https://cloud.githubusercontent.com/assets/1529481/15235318/cfe5c0e6-187f-11e6-9904-1cbdb51cbe7f.png)
2D Clustering:
![clustering_b0002](https://cloud.githubusercontent.com/assets/1529481/15235323/da3e076a-187f-11e6-8c70-d9a475b21743.png)
Populations used for standard curve generation:
[...]

Issue #212, opened 2016-05-16
DATA size does not match expected array size


agtebo:
I wanted to try out your software on my data, but I'm unable to load my files. The data was acquired on an Accuri c6 flow cytometer and are exported to FCS 3.0 format. Attempting to load my data using results in output: 
`ValueError: DATA size does not match expected array size (array size = 2800000 bytes, DATA segment size = 2800001 bytes)`

I'm pretty new to flow cytometry and I'm unsure how to troubleshoot this. Running OSX 10.11.4 and python 2.7.11 with Anaconda 4.0.0. I'm happy to share an example of my .fcs file if that would helpful. 

Thank you!

castillohair:
Hi,

We are working on an update to significantly improve compatibility (see #203). It should be coming out in the next couple of days. It would definitely be useful if you sent us your file for testing before we release the new version.

Thanks.

agtebo:
Cool! I apologize for submitting a duplicate issue, then. You can find an example .fcs from our instrument (Accuri c6) [here](https://app.box.com/s/g92de47hvqjlxbxuma49o2sdetd59g7f). This sample is bacteria expressing a fluorescent protein, nothing too fancy. Hope this helps.

castillohair:
Just wanted to give an update on this. The new version 1.1.0 has been released, which should be able to handle your file.

If you prefer writing code, the following script should read your file and produce a plot from it:

``` python
import FlowCal
from matplotlib import pyplot as plt
import numpy as np

# Load file
d = FlowCal.io.FCSData('A02 YFAST HMBR.fcs')

# Define channels
sc_channels = ['FSC-A', 'SSC-A']
fl_channels = ['FL1-A', 'FL2-A', 'FL3-A', 'FL4-A']

# Transform to a.u.
d = FlowCal.transform.to_rfi(d)

# Perform low-high gating
dg = FlowCal.gate.high_low(d, channels=sc_channels + fl_channels)

# Attempt density gating
dg, __, dc = FlowCal.gate.density2d(
    dg,
    channels=sc_channels,
    gate_fraction=0.75,
    full_output=True)

# Plot
FlowCal.plot.density_and_hist(
    d,
    gated_data=dg,
    gate_contour=dc,
    density_channels=sc_channels,
    density_params={'mode': 'scatter'},
    hist_channels=fl_channels,
    savefig='density_hist.png')
```

If you prefer using the Excel UI, I'm attaching an Excel file that should work with your file.

[...]

Issue #198, opened 2016-03-23
New example files


castillohair:
Our example files suck. We should get new ones.

castillohair:
The example should be an easy-to-understand dataset (i.e. IPTG dose-response curve). Files to be included are the following:
- An Excel file as input to the Excel UI.
- A script that shows how to MEF these flow files from the Python API.
- A script that shows how to process these flow files from the Python API in a.u.
- A script that shows how to process these flow files using the Excel file.

These three scripts could be regular python files (`.py`) or IPython notebooks. I favor the former because it does not introduce the additional dependency of IPython.

castillohair:
@JS3xton brought up a good point in #213. The example scripts in v1.0.0 use a small amount of code to accomplish their tasks, and do so by assuming certain things about the example flow files. For example, they don't check whether the files contain integer or floating point data, which affects whether `high_low` gating is adequate or not, Instead, they directly apply this gate using our knowledge of these files. This contrasts with the code in the Excel UI (`excel_ui.py`) in which everything is checked before performing most actions.

So there are two ways in which the new example scripts could go: We could do something similar to the Excel UI code and make code that works for the general case, or keep it narrow. I would argue that it is better to keep it narrow because that makes the code concise and easy to understand, and that is the goal of the example scripts - people are actually expected to look at them. However, I think comments in the script could do a better job at explaining each step, why each step is being applied, and when one should do something different.

JS3xton:
> better to keep it narrow

Agree. I think narrow is better (more likely to be read and more modularly informative). I think narrow can be justified if you set the stage appropriately (e.g. informative comments which describe additional info about a file (if the filename doesn't already communicate it) and why certain steps are performed (e.g. `high_low` gating)).

> regular python files (`.py`) or IPython notebooks

I keep going back and forth on this. I like that Jupyter notebooks can be rendered through GitHub, so if GitHub is the primary interface through which people are expected to interact with the examples, then I favor Jupyter notebooks. If, however, the examples will be more frequently consulted via the documentation website or in a cloned copy of the repo, then `.py` files sound simpler.

JS3xton:
[...]

Issue #79, opened 2015-10-16
Code style/formatting


JS3xton:
These stylistic/formatting things seem obvious to me, but multiple modules do not consistently conform to all of them:
* Function signatures
 * Parameters of multiline function signatures should be aligned vertically.

    [ref=[PEP 0008:Indentation](https://www.python.org/dev/peps/pep-0008/#indentation)]:

    > Continuation lines should align wrapped elements either vertically using Python's implicit line joining inside parentheses, brackets and braces, or using a hanging indent.
    >
    > Yes:
    >
    >        # Aligned with opening delimiter.
    >        foo = long_function_name(var_one, var_two,
    >                                 var_three, var_four)
 * Parameters of multiline function signatures should not have unnecessary trailing white spaces.
 * Input parameter defaults should have no spaces around the `=`. e.g. `param='default'`, not `param = 'default'`

    [ref=[PEP 0008](https://www.python.org/dev/peps/pep-0008/#other-recommendations)]:

    > Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.
    >
    > Yes:
    >
    >        def complex(real, imag=0.0):
    >            return magic(r=real, i=imag)
    >
    > No:
    >
    >        def complex(real, imag = 0.0):
    >            return magic(r = real, i = imag)

* New line continuation (`\`) should have space between last character and backslash (e.g. `palettable.colorbrewer \`, not `palettable.colorbrewer\`). This is a little more subjective, but it looks better to me this way, and the [PEP 0008:Maximum Line Length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) example does it this way.
* Math should generally have spaces between operators and variables unless compact grouping makes it more readable for long expressions (e.g. `y = m*x + b` instead of `y=m*x+b`).

    [ref=[PEP 0008](https://www.python.org/dev/peps/pep-0008/#other-recommendations)]:

    >If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.
    >
    > Yes:
    >
    >        i = i + 1
    >        submitted += 1
    >        x = x*2 - 1
    >        hypot2 = x*x + y*y
    >        c = (a+b) * (a-b)
    >
    > No:
    >
    >        i=i+1
    >        submitted +=1
    >        x = x * 2 - 1
    >        hypot2 = x * x + y * y
[...]

PR #213, opened 2016-06-02
MEFing for floating-point data

castillohair:
Solves #195.
- MEFing of data acquired with linear amplification is supported, both with the Excel UI and Python API. Changes required to support this include new or modified functions in `FlowCal.mef`, a new standard curve model, and the fact that all data should now be transformed manually to a.u. before MEF.
- Floating-point data with negative values is correctly processed. In particular, logicle scale has been implemented for visualization, and it is the default (Excel UI) for data acquired with linear amplification. Geometric statistics are calculated on the positive events, and a warning is thrown indicating the percentage of valid events.
- Tested with examples and private dataset aquired using the SEA's flow cytometer. The latter exports FCS files version 2.0 (integer) with either linear or log amplification, and FCS3.0 (floating-point, negative values) with linear amplification (log amplification with floating point is not allowed by the standard). MEFing with the Excel UI seems to be performed correctly, as far as I can tell, with these three datasets.


FlowCal/gate.py
```diff
@@ -275,13 +275,13 @@ def density2d(data,
...
+        ``log``, or ``logicle``. `xscale` is ignored in `bins is an array
+        or a list of arrays.
     yscale : str, optional
-        Scale of the bins generated for the y axis. Can be either
-        ``linear`` or ``log``. `yscale` is ignored in `bins is an array or
-        a list of arrays.
+        Scale of the bins generated for the y axis, either ``linear``,
+        ``log``, or ``logicle``. `yscale` is ignored in `bins is an array
+        or a list of arrays.
     sigma : scalar or sequence of scalars, optional
```
JS3xton, line 285: should the defaults for `xscale` and `yscale` for `fc.gate.density2d()` also be `logicle` now instead of `linear`?

FlowCal/io.py
```diff
@@ -1375,6 +1438,9 @@ def __new__(cls, infile):
         else:
             time_step = None
 
+        # Data type
+        data_type = fcs_file.text['$DATATYPE']
+
```
JS3xton, line 1444: might consider using `fcs_file.text.get()` instead of accessing the dictionary via a key. this is super minor, but the error behavior is different. `.get()` will just return `None` if the key is not present I believe; accessing the dictionary directly will throw a `KeyError`. I can see cases for both, but I think we started off using `.get()` and now appear to be inconsistent with our access

FlowCal/io.py
```diff
@@ -972,6 +976,17 @@ def analysis(self):
...
     @property
+    def data_type(self):
+        """
+        Type of data in the FCS file's DATA segment.
+
+        `data_type` is 'I' if the data type is integer, 'F' for floating
+        point, and 'D' for double.
+
+        """
+        return self._data_type
```
JS3xton, line 987: this `data_type` attribute is just for convenience? easier than making the user access the TEXT segment dictionary? was there a particular motivation for exposing this but not other things? (i don't really have strong feelings either way, just curious)

FlowCal/io.py
```diff
@@ -1243,36 +1258,68 @@ def resolution(self, channels=None):
         else:
             return self._resolution[channels]
 
-    def hist_bins(self, channels=None, nbins=None, scale='linear'):
+    def hist_bins(self, channels=None, nbins=None, scale='linear', **kwargs):
```
JS3xton: should the `scale` default be `logicle` I was under the impression we want to make that the new default everywhere (Excel UI, plotting function defaults, here, etc.)?

FlowCal/gate.py
castillohair, line 285: Actually, the default for functions has always been linear. What we discussed about logicle was for the Excel UI, as far as I understood. I'd rather not change this unless there is a good reason.

FlowCal/io.py
castillohair, line 987: There is no place in FlowCal in which we access `.text`, and I wanted to keep it that way.

FlowCal/io.py
castillohair: Excel UI only (see comment above).

FlowCal/io.py
castillohair, line 1444: Changed in 07857ddfd48dded01aac8321eee0a4f46a88667d.

FlowCal/gate.py
JS3xton, line 285: Oh, just for the same reasons that we went with logicle for the Excel UI. Seems more appropriate in the general case. And in an effort to try and standardize what kind of plot you're looking at. Makes sense to make one kind of plot the default everywhere. My feelings aren't super strong on this, though.

FlowCal/plot.py
```diff
@@ -64,6 +67,510 @@
...
+                M = 4.5
+            if W is None:
+                W = 0.5
+        # Check that property values are valid
+        if T <= 0:
+            ValueError("T should be positive")
+        if M <= 0:
+            ValueError("M should be positive")
+        if W < 0:
+            ValueError("W should not be negative")
```
JS3xton: should these errors be `raise`ed? or is this some black magic that i'm not familiar with? Same for the `ValueError` up at the beginning of `__init__()`

FlowCal/plot.py
```diff
@@ -64,6 +67,510 @@
...
+        # Functions to provide to the root finding algorithm
+        def W_f(p):
+            return 2*p / (p + 1) * np.log10(p)
+        def W_root(p, W_target):
+            return W_f(p) - W_target
+        # Find solution
+        sol = scipy.optimize.root(W_root, x0=p0, args=(W))
+        # Solution should be unique
+        assert sol.success
+        assert len(sol.x) == 1
```
JS3xton, line 319: Is there a reason to use `assert`s here? (as per discussion in #79, raising errors is preferred)

castillohair, line 319: I analyzed the equation extensively, and for `W > 0` there will always be one solution that should be relatively easy to find. This does not depend on user input. In fact, I cannot think how to make this fail. But if it does at some point, the assertion should make it easy to identify that this is the source of the problem. I believe this is consistent with https://wiki.python.org/moin/UsingAssertionsEffectively, particularly the third item in the following list:

> Places to consider putting assertions:
> - checking parameter types, classes, or values
> - checking data structure invariants
> - checking "can't happen" situations (duplicates in a list, contradictory state variables.)
> - after calling a function, to make sure that its return is reasonable

FlowCal/plot.py
castillohair: Lol corrected in 33dabbd8b98cac1034b6ca6eb38c8ef41e28aa84.

examples/analyze.py
```diff
@@ -39,38 +39,49 @@
...
+                                          sc_channels + fl_channels)
+
+    # Gating
+
+    # Remove first and last events. Transients in fluidics can make the
+    # first few and last events slightly different from the rest.
     beads_data = FlowCal.gate.start_end(beads_data, num_start=250, num_end=100)
+
+    # Remove saturating events in forward/side scatter.
     beads_data = FlowCal.gate.high_low(beads_data, sc_channels)
```
JS3xton, line 65: do you have specific knowledge of the files acted upon by `analyze.py` which necessitates `gate.high_low()`? In general, I don't think `gate.high_low()` is automatically warranted anymore (e.g. floating point data). might consider adding a check of the data type here unless you already know what these files are. sounded like the examples were gonna change soon anyways? so maybe not a big deal.

examples/analyze.py
```diff
@@ -39,38 +39,49 @@
...
     FlowCal.plot.density_and_hist(
         beads_data,
         gated_beads_data, 
         density_channels=sc_channels,
         hist_channels=fl_channels,
         gate_contour=gate_contour, 
-        density_params={'mode': 'scatter'}, 
-        hist_params={'ylim': (0, 1000)},
+        density_params={'mode': 'scatter', 'xscale': 'log', 'yscale': 'log'},
+        hist_params={'ylim': (0, 1000), 'xscale': 'log'},
```
JS3xton, line 84: did you go with `log` over `logicle` for a reason? again, figured we'd do `logicle` everywhere to be consistent. (same for all plots in this file)

examples/analyze.py
```diff
@@ -100,26 +111,32 @@
...
+
+    # Remove first and last events. Transients in fluidics can make the
+    # first few and last events slightly different from the rest.
+    data_gated = [FlowCal.gate.start_end(di, num_start=250, num_end=100)
+            for di in data_transf]
+
+    # Remove saturating events in forward/side scatter and fluorescence
+    # channels.
+    data_gated = [FlowCal.gate.high_low(di, sc_channels + mef_channels)
+            for di in data_gated]
```
JS3xton, line 133: same as above. if `gate.high_low()` is warranted for this particular set of example files, then this is fine. in the general case (as was done in the Excel UI I believe?), I think there should be a check of the data type

castillohair, line 133: Examples will be changed, so this does not matter too much. I would suggest moving this discussion to #198.

castillohair:
We need to have a deadline for the revision of this merge request. I will merge this on Sunday night if there are no further comments.

FlowCal/mef.py
```diff
@@ -182,37 +237,38 @@ def selection_std(populations,
...
     -------
     std_crv : function
-        Standard curve that transforms arbitrary fluorescence values from
-        channel units to MEF units. This function has the signature ``y =
-        std_crv(x)``, where `x` is some fluorescence value in channel units
-        and `y` is the same fluorescence expressed in MEF units.
+        Standard curve that transforms arbitrary fluorescence values to MEF
+        units. This function has the signature ``y = std_crv(x)``, where
+        `x` is some fluorescence value in arbitrary units and `y` is the
+        same fluorescence expressed in MEF units.
```
JS3xton: In description of standard curve, maybe more specifically mention that it transforms _RFI_ values to MEF units (not just "arbitrary fluorescence values"). I think channel values are arbitrary under all 3 cases (log amp + int data, lin amp + int data, float data)?, but this function no longer works as intended with the log amp + lin amp case. I think it also makes the serialization on `to_rfi()` clearer / more modular to explicitly state that at this point, we're only dealing with proper RFI values.

along these lines, maybe rebrand these variable names as `fl_rfi`?

FlowCal/mef.py
```diff
@@ -265,27 +320,34 @@ def fit_beads_autofluorescence(fl_channel, fl_mef):
...
     # 1: y-intercept found by putting a line through highest two points.
-    # 2: bead autofluorescence initialized to 100.
+    # 2: bead autofluorescence initialized using the first point.
     params[0] = (np.log(fl_mef[-1]) - np.log(fl_mef[-2])) / \
-                    (fl_channel[-1] - fl_channel[-2])
-    params[1] = np.log(fl_mef[-1]) - params[0] * fl_channel[-1]
-    params[2] = 100.
+                (np.log(fl_au[-1]) - np.log(fl_au[-2]))
+    params[1] = np.log(fl_mef[-1]) - params[0] * np.log(fl_au[-1])
+    params[2] = np.exp(params[0]*np.log(fl_au[0]) + params[1]) - fl_mef[0]
```
JS3xton: any reason you opted to update the initial guess for `params[2]`? I'll note that I suggested this particular update back in #165, but you didn't think it would matter. also, I don't have `- fl_mef[0]` at the end of the version I proposed. Is that usually just 0? but the version you have here is more general (in case the 1st peak is ever _not_ a blank)?

with that in mind, I still think my other suggestion proposed in #165 (initialize slope and y-intercept terms from a line fit with all the bead subpopulations, not just from the top two points. arguably more robust) is a good idea

FlowCal/mef.py
```diff
@@ -265,26 +319,30 @@ def fit_beads_autofluorescence(fl_channel, fl_mef):
...
 
-    # Channel-to-MEF standard curve transformation function
+    # a.u.-to-MEF standard curve transformation function
+    # We use the equivalent form ``exp(p[1]) * x**p[0]``. Mathematically, this
+    # is undefined for x < 0. In this case, we redefine the output as
+    # ``- exp(p[1]) * (-x)**p[0]``. These two can be combined as follows:
+    # ``np.sign(x) * np.exp(p[1]) * (np.abs(x)**p[0])``.
     def sc_fun(p,x):
-        return np.exp(p[0] * x + p[1])
+        return np.sign(x) * np.exp(p[1]) * (np.abs(x)**p[0])
```
JS3xton, line 417: i don't understand this comment. not sure what equation you are originally referring to, to which you claim `exp(p[1]) * x**p[0]` is equivalent. i imagine this is just a rearrangement of the bead model, minus the autofluorescence term? I also don't understand why it's undefined for x<0 (you can raise a negative number to an exponent...). do you mean arbitrary fluorescence is by definition nonzero (cause you can't have a _negative_ fluorescence)? in which case, why is it appropriate to just switch the sign when you _do_ encounter a negative arbitrary fluorescence value?

FlowCal/mef.py
```diff
@@ -308,38 +370,44 @@ def sc_fun(p,x):
...
     std_crv : function
-        The standard curve, mapping channel units to MEF units).
+        The standard curve, mapping arbitrary units to MEF units.
 
     Other parameters:
     -----------------
+    xscale : str, optional
+        Scale of the x axis, either ``linear`` or ``log``.
+    yscale : str, optional
+        Scale of the y axis, either ``linear`` or ``log``.
```
JS3xton, line 477: I feel like maybe the standard curve should have the option to be plotted on `logicle` scales to be consistent with the rest of the plots (again, attempting to unify the muscle memory / intuition we develop when analyzing flow data on plots generated by our package). haven't looked at a standard curve on those scales, though, so not sure if that is a bad idea for some unforeseen reason

JS3xton:
In light of us embracing the logicle scale, kind of a general thought regarding some functions in the `mef` module (originally raised regarding plotting the standard curve, but I'm realizing now that it extends to other parts of the module): should `mef.clustering_gmm()` and `mef.selection_std()` support `scale=logicle`? That sounds nontrivial, but if we don't support it, I guess it feels a little less clear to me where we're embracing logicle scale versus log scale versus linear scale.

FlowCal/mef.py
```diff
@@ -149,30 +168,66 @@ def selection_std(populations,
...
 
+    # Rescale if necessary
+    if scale=='log':
+        # Copy events
+        for i in range(len(populations)):
+            populations[i] = populations[i].copy()
+
+        # Log scale requires saturating negative and zero-valued events
+        for p in populations:
+            p[p < 1] = 1.
```
JS3xton: not sure what you mean by this comment. (shouldn't they already be saturated by the instrument? we're manually doing the same thing here?). Also, is `p<1` the right check? not sure what kind of data we have at this point, but with floating point, I imagine you could have valid (i.e. non-saturating) events between 0 and 1. Seems like the right check should be `p<=0`, with violators being replaced with something positive (which opens a new can of worms regarding what positive value should be used)

you sure it's not more appropriate to just discard values less than or equal to 0 when `scale=log`? instead of artificially designating them as `1.`

(seems like this change won't affect the behavior of this function much, though, so this might now matter much)

FlowCal/mef.py
```diff
@@ -71,6 +76,19 @@ def clustering_gmm(data,
...
+    # Check that ``scale`` is supported
+    if scale!='linear' and scale!='log':
+        raise ValueError("scale {} not supported".format(scale))
+
+    # Rescale if necessary
+    if scale=='log':
+        # Copy events
+        data = data.copy()
+        # Log scale requires saturating negative and zero-valued events
+        data[data < 1] = 1.
```
JS3xton: same comment as in `selection_std()` regarding clarity of comment and appropriateness of `data[data<1] = 1`

FlowCal/mef.py
```diff
@@ -71,6 +76,19 @@ def clustering_gmm(data,
...
+        # Rescale
+        data = np.log10(data)*256
```
JS3xton: where did the `*256` come from? that seems like an artifact from a previous (less general) version of this. if you want to include that scaling term, seems like this should really be transformed with the inverse of `to_rfi()` which would appropriately incorporate the amplifier resolution/range. that would go back to channel units, though. I think i'm in favor of just a standard natural log transformation on RFI values. (e.g. `np.log(data)`? where `data` should always be in RFI now?)

FlowCal/mef.py
```diff
@@ -149,30 +168,66 @@ def selection_std(populations,
...
+                low = 10**(np.log10(r[0]) + 0.015*(np.log10(r[1]/r[0])))
         else:
             raise TypeError("argument 'low' not specified")
     if high is None:
-        if hasattr(populations[0], 'range'):
-            high = 0.985*populations[0].range(0)[-1]
+        if hasattr(populations[0], 'hist_bins'):
+            if scale=='linear':
+                r = populations[0].range(0)
+                low = r[0] + 0.985*(r[1] - r[0])
```
JS3xton: looks like a copy-paste typo here. this `low` should be `high`? (it's under an if-else branch querying `high`)

FlowCal/mef.py
```diff
@@ -149,30 +168,66 @@ def selection_std(populations,
...
+        for p in populations:
+            p[p < 1] = 1.
+
+        # Rescale events
+        for i in range(len(populations)):
+            populations[i] = np.log10(populations[i])
+
+        # Rescale low and high thresholds
+        low = np.log10(low)
+        high = np.log10(high)
```
JS3xton: why are we doing `log10` instead of natural log? plotting appears to be on a log10 scale, I will grant you, but it seems like transformations on RFI values should be orthogonal to that and perhaps more simply or fundamentally transformed via natural log. this is probably not terribly important, though, and may skew what currently exists (what's the difference? a scaling factor of some kind I think?)

FlowCal/mef.py
```diff
@@ -149,30 +168,66 @@ def selection_std(populations,
...
-        and 0.0985 of the highest value in ``populations[0].range``.
+        `populations` are FCSData objects, use 1.5% and 98.5% of the range
+        in ``populations[0].range``.
     n_std_low, n_std_high : float, optional
         Number of standard deviations from `low` and `high`, respectively,
         that a population's mean has to be closer than to be discarded.
+    scale : str, optional
+        Rescaling applied to `populations` before calculating means and
+        standard deviations. If ``linear``, no rescaling is applied. If
+        ``log``, apply base-10 logarithm to `populations`.
```
JS3xton: seems like `scale` could be generalized to just accept a function (perhaps something conforming to the `transform` module function signatures?) which is applied to `data`, `low`, and `high`.

makes me wonder why this needs to be passed to this function explicitly in the first place, though, instead of just performing the transformation on `data` before calling this function. (this is a deeper seated debate we've had before which is kinda reflected in the `mef.get_transform_fxn()` function, as well as a few other places I think)

JS3xton: maybe change `scale` to `transform`, if the generalized transformation idea was embraced

FlowCal/mef.py
```diff
@@ -441,9 +510,9 @@ def get_transform_fxn(data_beads,
         Results of the subpopulation selection step, containing the
         following fields:
 
-        channel : list
+        au : list
             The fluorescence values of each selected subpopulation in
-            channel units, for each channel in `mef_channels`.
+            arbitrary units, for each channel in `mef_channels`.
```
JS3xton: again, maybe rebrand as `rfi`? to be more precise / clear that we should only be dealing with RFI values at this point

FlowCal/mef.py
```diff
@@ -547,9 +616,10 @@ def get_transform_fxn(data_beads,
     2. The fluorescence of each subpopulation is calculated, for each
        channel in `mef_channels`.
     3. Some subpopulations are then discarded if they are close to either
-       the minimum or the maximum channel value. In addition, if the MEF
-       value of some subpopulation is unknown (represented as a ``np.nan``
-       in `mef_values`), the whole subpopulation is also discarded.
+       the minimum or the maximum cannel range limits. In addition, if the
```
JS3xton: typo ("cannel")

FlowCal/mef.py
```diff
@@ -631,21 +705,30 @@ def get_transform_fxn(data_beads,
...
             plt.figure(figsize=(6,4))
-            FlowCal.plot.scatter2d(populations,
-                                   channels=clustering_channels,
-                                   savefig=savefig)
+            FlowCal.plot.scatter2d(
+                populations,
+                channels=clustering_channels,
+                xscale='log' if clustering_amp_type[0][0] else 'logicle',
+                yscale='log' if clustering_amp_type[1][0] else 'logicle',
+                savefig=savefig)
```
JS3xton, line 801: i don't remember when `clustering_amp_type[0][0]` evaluates to `True` (linear vs. log amp?). would be nice if the code could somehow convey that. otherwise, specifying in comments might help.

FlowCal/mef.py
```diff
@@ -654,6 +737,9 @@ def get_transform_fxn(data_beads,
             FlowCal.plot.scatter3d_and_projections(
                 populations,
                 channels=clustering_channels[:3],
+                xscale='log' if clustering_amp_type[0][0] else 'logicle',
+                yscale='log' if clustering_amp_type[1][0] else 'logicle',
+                zscale='log' if clustering_amp_type[2][0] else 'logicle',
```
JS3xton: same as above: i don't remember when `clustering_amp_type[0][0]` evaluates to `True`

FlowCal/mef.py
JS3xton, line 801: on this note, are you using `scale=log` if the amplifier is `log`? I'm not sure I think that's necessarily warranted. Even if data was collected on a log amplifier, I think data should always be displayed on logicle scale unless explicitly specified otherwise by the user; display scale should be orthogonal to amplifier scale

FlowCal/mef.py
```diff
@@ -631,21 +705,30 @@ def get_transform_fxn(data_beads,
...
             plt.figure(figsize=(8,4))
-            FlowCal.plot.hist1d(populations,
-                                channel=clustering_channels[0],
-                                bins=256,
-                                alpha=0.75,
-                                savefig=savefig)
+            FlowCal.plot.hist1d(
+                populations,
+                channel=clustering_channels[0],
+                xscale='log' if clustering_amp_type[0][0] else 'logicle',
```
JS3xton: same as below: i don't remember when `clustering_amp_type[0][0]` evaluates to `True`

FlowCal/excel_ui.py
```diff
@@ -216,7 +217,8 @@ def process_beads_table(beads_table,
     This function processes the entries in `beads_table`. For each row, the
     function does the following:
         - Load the FCS file specified in the field "File Path".
-        - Transform the forward scatter/side scatter channels if needed.
+        - Transform the forward scatter/side scatter and fluorescence
+          channels to a.u.
```
JS3xton: maybe transform to _RFI_ (same vein as other comments: more precise and enforcing serialization on RFI)

FlowCal/excel_ui.py
```diff
@@ -357,57 +360,78 @@ def process_beads_table(beads_table,
...
-                xscale='log'
-                       if beads_sample_gated.amplification_type(
-                           sc_channels[0])[0]
-                       else 'linear',
-                yscale='log'
-                       if beads_sample_gated.amplification_type(
-                           sc_channels[1])[0]
-                       else 'linear',
+                xscale='log' if sc_amp_type[0][0] else 'logicle',
+                yscale='log' if sc_amp_type[1][0] else 'logicle',
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier.

FlowCal/excel_ui.py
```diff
@@ -357,57 +360,78 @@ def process_beads_table(beads_table,
...
                 density_params["title"] = "{} ({:.1f}% retained)".format(
                     beads_id,
                     beads_sample_gated.shape[0] * 100. / beads_sample.shape[0])
+                # Axis scaling will be logarithmic if data was acquired with a
+                # log amplifier, and logicle otherwise.
+                sc_amp_type = beads_sample_gated.amplification_type(sc_channels)
+                density_params['xscale'] = \
+                    'log' if sc_amp_type[0][0] else 'logicle'
+                density_params['yscale'] = \
+                    'log' if sc_amp_type[1][0] else 'logicle'
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier.

FlowCal/excel_ui.py
```diff
@@ -357,57 +360,78 @@ def process_beads_table(beads_table,
...
+                                   95) * (10**0.75),
+                     )
+                # Histogram plot parameters
+                # Axis scaling will be logarithmic if data was acquired with a
+                # log amplifier, and logicle otherwise.
+                cluster_amp_type = \
+                    beads_sample_gated.amplification_type(cluster_channels)
+                hist_params = \
+                    [{'xscale': 'log' if ch_amp_type[0] else 'logicle'}
+                     for ch_amp_type in cluster_amp_type]
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier.

FlowCal/excel_ui.py
```diff
@@ -489,14 +479,14 @@ def process_samples_table(samples_table,
     The function processes each entry in `samples_table`, and does the
     following:
         - Load the FCS file specified in the field "File Path".
-        - Transform the forward scatter/side scatter channels if needed.
+        - Transform the forward scatter/side scatter to a.u., if needed.
```
JS3xton: same as above: "transform to _RFI_" to be more precise

FlowCal/excel_ui.py
```diff
@@ -689,7 +679,9 @@ def process_samples_table(samples_table,
                                         beads_dv,
                                         sample.detector_voltage(fl_channel)))
 
-                        # Attempt to transform
+                        # Transform first to RFI
```
JS3xton: not sure what "first" is referring to

FlowCal/excel_ui.py
```diff
@@ -736,51 +764,55 @@ def process_samples_table(samples_table,
...
                 channels=sc_channels,
                 gate_fraction=sample_row['Gate Fraction'],
-                xscale='log'
-                       if sample_gated.amplification_type(sc_channels[0])[0]
-                       else 'linear',
-                yscale='log'
-                       if sample_gated.amplification_type(sc_channels[1])[0]
-                       else 'linear',
+                xscale='log' if sc_amp_type[0][0] else 'logicle',
+                yscale='log' if sc_amp_type[1][0] else 'logicle',
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle, even with a log amplifier.

FlowCal/excel_ui.py
```diff
@@ -736,51 +764,55 @@ def process_samples_table(samples_table,
...
                     sample_id,
                     sample_gated.shape[0] * 100. / sample.shape[0])
-                # Define histogram plot parameters
+                # Axis scaling will be logarithmic if data was acquired with a
+                # log amplifier, and logicle otherwise.
+                sc_amp_type = sample_gated.amplification_type(sc_channels)
+                density_params['xscale'] = \
+                    'log' if sc_amp_type[0][0] else 'logicle'
+                density_params['yscale'] = \
+                    'log' if sc_amp_type[1][0] else 'logicle'
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier.

same as mentioned elsewhere: don't remember when `sc_amp_type[0][0]` evaluates to `True`. your comments explain it better here, so perhaps this is good enough. would be nice if code explained itself, though

FlowCal/excel_ui.py
```diff
@@ -736,51 +764,55 @@ def process_samples_table(samples_table,
...
-                        param['xscale'] = 'log'
-                    else:
+                    # Only channel numbers are plotted in linear scale
+                    if (ru == 'Channel Number'):
                         param['xscale'] = 'linear'
+                    else:
+                        # Axis scaling will be logarithmic if data was acquired
+                        # with a log amplifier, and logicle otherwise.
+                        rc_amp_type = sample_gated.amplification_type(rc)
+                        param['xscale'] = 'log' if rc_amp_type[0] else 'logicle'
```
JS3xton: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier.

same as mentioned elsewhere: don't remember when `rc_amp_type[0][0]` evaluates to `True`. your comments explain it better here, so perhaps this is good enough. would be nice if code explained itself, though

FlowCal/plot.py
```diff
@@ -917,8 +1000,8 @@ def density2d(data,
             cbar.ax.set_ylabel('Counts')
 
     # Set scale of axes
-    plt.gca().set_xscale(xscale)
-    plt.gca().set_yscale(yscale)
+    plt.gca().set_xscale(xscale, data=data_plot, channel=0)
+    plt.gca().set_yscale(yscale, data=data_plot, channel=1)
```
JS3xton, line 1012: kinda weird that you pass additional parameters to a `set_scale()` function, but I can see its usefulness. these keywords just get passed to the scale object's constructor, though, right? do we know that these particular parameter names won't clobber/overload anything in the native matplotlib scale objects?

FlowCal/plot.py
```diff
@@ -1058,21 +1141,33 @@ def scatter2d(data_list,
...
-            hasattr(data_plot.hist_bins, '__call__'):
-        plt.ylim(data_plot.hist_bins(channels=1, nbins=1, scale=yscale))
+    if xlim is None:
+        xlim = [np.inf, -np.inf]
+        for data in data_list:
+            if hasattr(data, 'hist_bins') and \
+                    hasattr(data.hist_bins, '__call__'):
+                xlim_data = data.hist_bins(channels=channels[0],
+                                           nbins=1,
+                                           scale=xscale)
```
JS3xton, line 1170: what's the difference between `hist_bins(..., nbins=1)` and getting the range?

FlowCal/plot.py
```diff
@@ -735,38 +1266,53 @@ def scatter3d(data_list,
...
-            data_plot_z = np.log10(data_plot[:,2])
-        else:
-            raise ValueError('scale {} not supported'.format(zscale))
         # Make scatter plot
-        ax_3d.scatter(data_plot_x,
-                      data_plot_y,
-                      data_plot_z,
+        ax_3d.scatter(xscale_transform(data_plot[:, 0]),
+                      yscale_transform(data_plot[:, 1]),
+                      zscale_transform(data_plot[:, 2]),
```
JS3xton, line 1328: looks like the 3d scatter plot loses context for the transform (and resulting scale) of each dimension which results in contextless axes (standard linear ticks; doesn't acknowledge logicle or log scale when appropriate. no labels of orders of magnitude, etc). would be kinda nice if the 3d scatter plot was consistent with the other 2d plots regarding the axes, but this may be a lot of work for what is otherwise a very minor aesthetics issue.

FlowCal/excel_ui.py
```diff
@@ -1126,23 +1181,24 @@ def generate_histograms_table(samples_table, samples, max_bins=1024):
...
+                # Decide which scale to use
+                # Channel units result in linear scale. Otherwise, scale is
+                # decided based on amplification type.
+                if unit == 'Channel':
+                    scale = 'linear'
+                elif sample.amplification_type(channel)[0]:
                     scale = 'log'
                 else:
-                    scale = 'linear'
+                    scale = 'logicle'
```
JS3xton, line 1168: same as mentioned elsewhere: I think amplifier scale and plotting (and gating) scale are orthogonal, so I would still go with logicle gating, even with a log amplifier. if `unit` is "Channel", though, then I agree with what @castillohair has said elsewhere that user probably wants raw data, and I think linear scale is appropriate for that.

same as mentioned elsewhere: don't remember when `sample.amplification_type(channel)[0]` evaluates to `True`. comments and code imply a relationship, but would be nice if it was more explicit

FlowCal/excel_ui.py
```diff
@@ -1126,23 +1181,24 @@ def generate_histograms_table(samples_table, samples, max_bins=1024):
...
-                    bin_centers = (np.log10(bin_edges[:-1]) + \
-                        np.log10(bin_edges[1:]))/2
-                    bin_centers = 10**bin_centers
+                # Calculate bin edges and centers
+                # We generate twice the necessary number of bins. We then take
+                # every other value as the proper bin edges, and the remaining
+                # values as the bin centers.
+                bins_extended = sample.hist_bins(channel, 2*nbins, scale)
+                bin_edges = bins_extended[::2]
+                bin_centers = bins_extended[1::2]
```
JS3xton, line 1177: this method for populating bin centers / edges means the centers are not what you might expect (Euclidean midpoint between the two bin edges). I think the method is fine, just might want to make sure you describe it clearly wherever it gets documented.

FlowCal/excel_ui.py
```diff
@@ -1054,18 +1083,44 @@ def add_samples_stats(samples_table, samples):
...
+                    channel + ' Geom. Mean',
+                    FlowCal.stats.gmean(sample_positive, channel))
+                samples_table.set_value(
+                    row_id,
+                    channel + ' Geom. Std',
+                    FlowCal.stats.gstd(sample_positive, channel))
+                samples_table.set_value(
+                    row_id,
+                    channel + ' Geom. CV',
+                    FlowCal.stats.gcv(sample_positive, channel))
```
JS3xton, line 1101: does processing the geometric statistics at the end of `add_samples_stats()` change the order in which the statistics are reported in the XLSX sheet? I wasn't particularly tied to a previous order, but just wanted to check if this was done on purpose.

FlowCal/excel_ui.py
```diff
@@ -1054,18 +1083,44 @@ def add_samples_stats(samples_table, samples):
...
+                    # Throw warning
+                    msg = "Geometric statistics for channel" + \
+                        " {} calculated on positive events".format(channel) + \
+                        " only ({:.1f}%). ".format(
+                            100.*sample_positive.shape[0]/sample.shape[0])
+                    warnings.warn("On sample {}: {}".format(row_id, msg))
+                    # Write warning message to table
+                    if samples_table.loc[row_id, 'Analysis Notes']:
+                        msg = samples_table.loc[row_id, 'Analysis Notes'] + msg
+                    samples_table.set_value(row_id, 'Analysis Notes', msg)
```
JS3xton, line 1086: we've talked previously about accumulating warnings throughout the duration of a particular analysis run. here, you seem to simply rely on concatenating warning strings (as opposed to a more explicit data structure for carrying around warnings)? is that correct? And you've padded all your warnings with a trailing space so they can concatenate in a general fashion?

FlowCal/plot.py
```diff
@@ -64,6 +67,511 @@
...
+                for d in data:
+                    # Extract channel
+                    y = d[:, channel] if d.ndim > 1 else d
+                    if hasattr(y, 'range') and hasattr(y.range, '__call__'):
+                        Ti = y.range(0)[1]
+                    else:
+                        Ti = np.max(y)
+                    T = Ti if Ti > T else T
+            if M is None:
+                M = 4.5 / np.log10(262144) * np.log10(T)
```
JS3xton: you originally had `M = 4.5` (which I think came straight from the Logicle paper?), but added this scaling factor. `np.log10(262144)` also comes from the paper? and you've just made this scale to the data a little better by effectively incorporating a ratio of `T` (extracted as max value from data?) and the max value used in the paper (262144)? is this right?

FlowCal/mef.py
```diff
@@ -265,27 +320,34 @@ def fit_beads_autofluorescence(fl_channel, fl_mef):
...
+        return np.sign(x) * np.exp(p[1]) * (np.abs(x)**p[0])
     
     # Fit parameters
-    err_par = lambda p: err_fun(p, fl_channel, fl_mef)
-    res = minimize(err_par, params, options = {'gtol': 1e-6})
+    err_par = lambda p: err_fun(p, fl_au, fl_mef)
+    res = minimize(err_par,
+                   params,
+                   bounds=((None, None), (None, None), (0, None)),
+                   options = {'gtol': 1e-10, 'ftol': 1e-10})
```
JS3xton, line 424: how come you upped the tolerances? don't really have a problem with it, just wondering if it was necessitated by something?

examples/analyze.py
castillohair, line 65: See #198.

FlowCal/mef.py
castillohair: The new equation is a better estimate for `params[2]`.

(In reality, I changed this because the fitting function was not working at some point, and I thought this was the problem. Turns out I had a typo somewhere else, but then I decided to leave the new estimate).

If fluorescence of the populations looks as we expect, I think using the last two peaks is better than   fitting a line. If fluorescence is different than we expect, you might be right that fitting a line is more robust. However, all the test data that I have show that fitting is working well. Changing this should be a new issue, and I could give you all the test data if you want to show that this change will improve things. I'm not particularly interested as I don't think this is a problem, though.

FlowCal/mef.py
castillohair, line 417: >  i imagine this is just a rearrangement of the bead model, minus the autofluorescence term?

Yes.

> you can raise a negative number to an exponent...

Not really... http://math.stackexchange.com/questions/415141/is-there-a-problem-when-defining-exponential-with-negative-base. In fact, the reason why I used this "extended" equation is because the numpy functions involved were initially failing for negative values.

I guess a more formal way to do it is the following. If `x<0`:

```
x**p0 = exp(Log(x))**p0
x**p0 = exp(Log(abs(x)) + i*pi)**p0
x**p0 = (exp(log(abs(x))) * exp(i*pi))**p0
x**p0 = (exp(log(abs(x))) * exp(i*pi))**p0
x**p0 = (abs(x)**p0) * (-1)**p0
```

Where I'm using the definition of complex logarithm (https://en.wikipedia.org/wiki/Complex_logarithm). In the last equation, `(-1)**po` is still not well defined, so I can't think of a 100% formally correct solution. Equivalently,

```
x**p0 = (abs(x)**p0) * exp(i*p0*pi)
```

I'm relying on `p0` being close to one (standard curve is close to a line) to approximate this as `exp(i*pi) = -1`. If it's not one, there might be some distortion applied to negative values because `exp(i*p0*pi)` is complex. We can still justify this equation by saying that we're using the negative absolute value of `exp(i*p0*pi)` (which is `-1`). Using the negative guarantees that negative numbers in a.u. remain negative. If we choose to use the real part of `exp(i*p0*pi)`, i.e. `cos(p0*pi)`, this value will be between `-1` and `1`, and I don't think this is appropriate.

Again, this method relies on `p0` being close to 1. This is not perfect, but all the floating-point data in my test dataset has given a `p0` that is very close to one, up to 1e-2 error. If you have a better idea on how to handle this, let me know. What we **can't** do is ignore or discard negative values. This is why we have this code review stage, as I didn't anticipate this issue until I started writing code.

If this makes sense, I'll update the comments to reflect this (better) description.

FlowCal/mef.py
castillohair, line 477: Made a quick attempt at this, and these are some resulting plots:

Integer data:
std_crv_fl1_b0002

Floating-point data
std_crv_apc-a_b3

When I look at the standard curves, I try to quickly identify how many decades in a.u. are mapped to how many decades in MEF. These plots don't help me do that. In fact, I think the shape distortion associated with the logicle scale makes a lot of information not easily observable.

Another argument for keeping this plot in log scale is that the standard curve is a line in log scale, so this line should be easier to see if the plot itself is in log scale.

Finally, this function is not supposed to be accessed even by an API user, and instead used by `get_transform_fxn`. It is also not in the plot module, where I would agree it is a good idea to be consistent and make all functions work with logicle.

For all these reasons, I don't think it is a good idea to make this plot in logicle.

FlowCal/mef.py
castillohair: Parameters `tol` and `min_covar` required extensive tuning for this function to work. If I change the scaling, I'll have to retune these parameters. That's a lot of effort to change something that already works well.

FlowCal/mef.py
castillohair: Yep typo. Fixed in fd93da2b574cb7558c833290ea7a55e0fef6fbe5.

FlowCal/mef.py
castillohair: Not only plotting, but in bin generation (which is used in plotting and gating, among others) `scale=log` is assumed to be `log10` and not `log`. I'd rather keep this consistent.

FlowCal/mef.py
castillohair: This could be another issue. Don't think is worth doing now, though.

FlowCal/mef.py
castillohair: Fixed in 7ca014e1a88b385cd26b876acb6070a790fa32db.

FlowCal/plot.py
castillohair, line 1012: > these keywords just get passed to the scale object's constructor, though, right?

Yes.

> do we know that these particular parameter names won't clobber/overload anything in the native matplotlib scale objects?

yep. check this: http://matplotlib.org/api/axes_api.html#matplotlib.axes.Axes.set_xscale

FlowCal/plot.py
castillohair, line 1328: Axes tick values are not shown in this function by default, so this is not terribly important.

> this may be a lot of work for what is otherwise a very minor aesthetics issue.

Exactly. This can be a future issue, though.

FlowCal/excel_ui.py
castillohair, line 1177: Not sure I understand this. I'm not claiming that the "centers" are actually the euclidean midpoint.

FlowCal/excel_ui.py
castillohair, line 1086: Yep.

FlowCal/excel_ui.py
castillohair, line 1101: Order is maintained, due to the instructions in lines 1037-1048.

FlowCal/plot.py
castillohair: Yep. Issue #212 included some files that had a range up to 1e7, and the plot didn't look very good with M=4.5. I think the paper was assuming that all floating-points had a range up to 262144, but that has changed.

FlowCal/mef.py
castillohair, line 424: I tested different tolerances and saw that the parameters were not converging completely with the previous set of tolerances, and with the new constrained method (note that we're indicating `minimize` that p[2] can't be negative, and so it uses a different constrained optimization method).

FlowCal/plot.py
castillohair, line 1170: `hist_bins` automatically handles the case in which `log` scaling is requested, but the lower limit of `range` is zero (you can't log zero).

Another way to handle this would be to get the range, explicitly check if the lower value is zero, and take appropriate action. But this happened frequently enough that I though using `hist_bins`  in all these instances was a better option.

FlowCal/mef.py
castillohair: > you sure it's not more appropriate to just discard values less than or equal to 0 when scale=log

The clustering function, at least, needs all the events even if they're outside of range.

I'll try adding logicle to this function, in which case saturation will not be necessary.

FlowCal/excel_ui.py
castillohair: Added comments about `rc_amp_type` in 3aa9551ffd27606183fd28ac7d7122ba3c1ebdc1.

FlowCal/excel_ui.py
castillohair: Added comments about `sc_amp_type` in 3aa9551ffd27606183fd28ac7d7122ba3c1ebdc1.

FlowCal/mef.py
castillohair: Added comments about `clustering_amp_type` in 3aa9551ffd27606183fd28ac7d7122ba3c1ebdc1.

FlowCal/excel_ui.py
castillohair: Changed in ad1298fa140baa5d868197b2f84467676e4d705f

FlowCal/excel_ui.py
castillohair: Changed in ad1298fa140baa5d868197b2f84467676e4d705f.

FlowCal/mef.py
castillohair: Changed in ad1298fa140baa5d868197b2f84467676e4d705f.

FlowCal/mef.py
castillohair: Changed in ad1298fa140baa5d868197b2f84467676e4d705f.

FlowCal/mef.py
castillohair, line 801: Added comments about clustering_amp_type in 3aa9551.

See below for discussion on scales.

FlowCal/mef.py
castillohair: Added comments about clustering_amp_type in 3aa9551.

FlowCal/excel_ui.py
castillohair: Clarified in f6b51c1a64162bcfde0f3dfb281d935c77a00703

castillohair:
There are several comments about using logicle scale independently of the amplifier type. Here I present my rationale for not doing so.

Here is a sample from the example files (S0008), plotted in log and logicle scale.
Log:
s0008_log

Logicle:
s0008_logicle

From here, one can see that the logicle plot results in a more jagged histogram shape. This has been consistent in all the integer, log samples that I've examined. I think this has an explanation. Because samples in log scale are required to be stored as integers, they have a finite resolution. If the resolution is divisible by the number of bins chosen, using log scale guarantees that each "display" bin contains the same number of "data" bins (e.g. with a 10 bit ADC, we have 1024 data bins, and we can use 1024, 512, 256.. etc. bins for plotting). There is no way to achieve the same with logicle. On the other hand, we have seen that with floating-point data the actual ADC resolution is often not properly stored, or it's variable, or it is very high. So it is harder to predict what bins to use, or it is not that important. On top of that we have negative numbers. This results in logicle being more appropriate for this particular case.

This is not really a great issue in my opinion. I thought that using logicle everywhere was going to be better because of consistency. But looking at the plots made me change my mind. The slight aesthetic improvement in my opinion is more important than "consistency" in this case. And also note that this does not affect processing or calculations, it only affects plots.

castillohair:
A related issue is the default scaling in gating and plotting function signatures. I can see an argument for changing this to `logicle`, and I'm not opposed to it. But I just wanted to point out that in the previous version the default was `linear` and not `log`, and the purpose of `logicle` is to replace `log` when negative events exist. So we can either keep it `linear` and be consistent with the previous version, change it to `logicle`, or allow a value of `None`, such that the relevant functions decide whether to use `log` or `logicle` based on the amplification type, the same way as the Excel UI is doing now.

FlowCal/plot.py
castillohair, line 1170: Yet another way would be to make the range go from `1` to `$PnR` instead of `0` to `$PnR - 1`. We talked about this in #196, where I said the following:

> Assume that the range is `[0, $PnR]` and the resolution is `$PnR`. The current implementation of `hist_bins()` would then generate, by default, `$PnR` bins with weird spacing different than one. We could live with these weird bins, or we could make the resolution `$PnR + 1` with floating points, or we could have the range always be `[0, $PnR - 1]`.
> 
> I choose the latter. So far we have two independent examples of floating point files, one of which blatantly disregards the standard's description of `$PnR` (i.e. has obviously a higher resolution ADC). In this file, events seem to be in `[0, $PnR]`, although there are negative numbers. Our other, more standard example does have channels with events in `[0, $PnR - 1]`, while using an ADC that clearly has the resolution specified by `PnR`, as I showed you before.
> 
> Ultimately, the precise range for floating points may not be that important, given that the standard specifies that the file may contain numbers outside it. I think what I chose is the best option, but I don't think it's better by much.

However, if we need to plot in log scale, having the lower limit be zero is inconvenient, because `log(0)` is undefined. So one solution would be to change this to `[1, $PnR]`, and then we would be able to always call `.range` and log the result.

castillohair:
So the issues in which I'd still like feedback are the following:
- Model's equations, particularly about the way we're applying the standard curve to negative events.
- Using `hist_bins(..., nbins=1)` in cases where actually using the `range` is to be expected. As I said, using the range directly is problematic for log scale because if the lower limit is zero, its logarithm is undefined. `hist_bins` takes care of this automatically, because it gets knowledge of the relevant scale via its `scale` parameter. But perhaps there is a better way to do this.
- The scales used in the Excel UI and default scales in functions.

I'm gonna try to do `logicle` in `mef.clustering_gmm` and `mef.selection_std`, and see how that looks.

FlowCal/mef.py
castillohair: Apparently changing the `min_covar` parameter from `1e-2` to `5e-5` (rounded from `1e-2/256 ~= 3.9e-5`) was enough to compensate for removing the `*256`. 352b8f4d47952e4e7c861e69c1e2d9e1b283aca5 implements this.

FlowCal/mef.py
JS3xton: saw one instance where it looked like you accidentally replaced `fl_au` with `fl_mef` (instead of `fl_rfi`). noted it in a separate comment.

castillohair:
As for 1af6868f8fe3c9c352c169af05e8bedff2e8931c, logicle scaling has been implemented in `mef.clustering_gmm` and `mef.selection_std`. Also, I changed my mind about the default scaling, and everything now defaults to `logicle`. While doing this, some small improvements have been made in `clustering_gmm`.

I have tested for correctness on most of my test dataset. I'll keep testing later today, but otherwise I think the code is ready to merge, as far as I can tell.

castillohair:
`fl_rfi` typo corrected in b58ef6412d1d7399dcb900cf9842e4fc977ecdd2.

castillohair:
I'll merge tomorrow morning if there is nothing else.

FlowCal/mef.py
JS3xton: I know our cytometer misbehaves with peaks up near the top end (you can sometimes see an explicit kink in the standards curve with the last peak if you don't manually throw it out), which undermines my faith in the top two peaks in general. That being said, usually that top peak should be thrown out if necessary by a user who is paying attention, and I still think the fitting winds up at the same solution, so I think you're probably right that the end result is good enough.

FlowCal/mef.py
JS3xton, line 417: Might consider clarifying in the comment that `exp(p[1]) * x**p[0]` is only undefined for `x < 0` _and_ non-integer `p[0]` (I think? perhaps a more precise set of numbers for `p[0]` for which that's undefined. It's still well defined for integer values of `p[0]`, though, I think, which is where I was getting hung up).

This all arises from this poorly justified `p[0]` slope parameter, which I still feel uneasy about. You're kinda arguing both sides here where you say it's consistently super close to 1 to justify this "extended" equation, but it's not consistently close enough to get rid of the parameter entirely.

I'm still trying to wrap my head around what exactly this is doing in the complex plane and how this would distort data if `p[0]` deviates more significantly from 1.0. Will come back to this...

FlowCal/mef.py
JS3xton, line 477: > When I look at the standard curves, I try to quickly identify how many decades in a.u. are mapped to how many decades in MEF.

I admittedly _don't_ try and identify how many decades in a.u. are mapped to how many decades are in MEF. I feel like you may be doing that due to your extensive exploration into the differences among amplifier (log, lin) and data (int, float) types.

Hadn't really thought about what I look for in a standard curve, but I guess it's (1) how closely the beads model fits the bead data points and (2) how significant the deviation on the low end is due to autofluorescence.

With those things in mind, this view definitely looks weird to me. I'm also confused why there aren't negative values on the top plot. Do you only show negative numbers if necessary? I'm struggling to identify where the linear-to-log transition occurs on the scales, too. These look just like log scales; is that just cause this was a quick attempt so the axes aren't right? or am I missing something?

> Another argument for keeping this plot in log scale is that the standard curve is a line in log scale, so this line should be easier to see if the plot itself is in log scale.

The standard curve should be a line on a _linear_ scale now (RFI -> MEF) (almost... stupid `p[0]`... we even claimed in Note 3 that RFI and MEF are both supposed to be proportional to photons... boooo). Indeed, on a log-log plot, that should also still be a line, but I would push it even further and say that perhaps a _linear_ scale would naively be most appropriate. In practice, we've discussed that it's more appropriate to go to a log space because are beads are logarithmically distributed in their fluorescence. I think because of _that_ reason, you can justify a log plot view of the standard curve. If we were using linearly distributed beads, though, I think I would say a linear-linear plot would be most appropriate. This again all boils down to the fact that displaying this standard curve (which is _almost_ a linear mapping from RFI -> MEF) and visualizing cell populations (which vary over many orders of magnitude but also have data near or below 0) are two fundamentally distinct visualization tasks, so it is not inconsistent to show one on log axes and the other on logicle axes.

> Finally, this function is not supposed to be accessed even by an API user, and instead used by get_transform_fxn. It is also not in the plot module, where I would agree it is a good idea to be consistent and make all functions work with logicle.

meh. I think these are weaker arguments. any function should still work as a stand-alone, in my opinion. And i think the plotting for the entire package should be consistent; nothing special about being in the `plot` module.

All this being said, I think I agree, though, that I wanna keep it as it was (log scale).

FlowCal/plot.py
JS3xton, line 1012: ahh perfect. good to see that other existing scales already use kwargs. excellent extension of that

FlowCal/plot.py
JS3xton, line 1170: I guess I feel like I'm OK with cleverly generating 1 large hist bin via `hist_bins()` to get the lower limit. Would just be nice to answer my initial question in the code via a comment, though (because this particular implementation really does beg that question).

I think setting the range to start at 1 would be cleaner for this particular use case (less code, less explanation). Not sure if that's a good idea everywhere, though. Don't remember all the places where range is used or why exactly we went with 0 as the bottom in the first place. If we went with `[1,$PnR]`, would the 0 bin get put with the 1 bin for integer data and $PnR bins generated (i.e. a full resolution histogram for integer data)? And we don't care about data less than 0 because we have logicle display now? Does that data just get thrown out if you manually specify log? (I think you should still be able to plot on log scales with negative data have it either gracefully fail or plot some intuitive subset).

Ultimately I think I'll defer to what you think is best here.

FlowCal/excel_ui.py
JS3xton, line 1177: sorry, I'm thinking from the perspective of a user of the Excel UI. Bin centers are only ever calculated to output into the Histogram tab of the Excel UI, right? I'm thinking from that perspective as someone reading "Bin Centers" in the Histogram tab (not from a coder reading this block of code); they may expect "Bin Centers" to indicate the Euclidean midpoint. still think this method is fine, just that you might want to make sure it's clearly described to the end user who's using "Bin Centers"

JS3xton:
Would be nice to see what the final product looks like (mostly relevant for the new logicle plots) across the 3 amplifier-datatype use cases (lin vs. log; int vs. float). Maybe that's the same as what's in #195? can't tell

FlowCal/excel_ui.py
```diff
@@ -403,6 +404,9 @@ def process_beads_table(beads_table,
                      np.percentile(beads_sample_gated[:, sc_channels[1]],
                                    95) * (10**0.75),
                      )
+                # Beads have a tight distribution, so less smoothing should be
+                # applied for visualization
+                density_params['sigma'] = 5.
```
JS3xton, line 409: this is specific to beads? I was originally gonna say that the default on `plot.density2d()` should just be updated, but you think the current default (`sigma=10`?) is still appropriate for cells?

FlowCal/io.py
JS3xton: gonna bump this, in light of all the discussion. should the default here for `hist_bins()` be `logicle`? or do you still think it should be `linear`, in spite of the embrace of logicle everywhere else? not sure if this snuck through un-updated or if this was intentionally left linear.

examples/analyze.py
JS3xton, line 84: ignoring this because it's `analyze.py` which is gonna get updated in #198

FlowCal/mef.py
JS3xton, line 417: Thought about this some more, and it actually kinda points out the usefulness of a logicle plot for the standard curve.

At the end of the day, we are extending the standard curve beyond 0; it would be nice to visualize that extension (i.e. plot the full standard curve, sufficiently extended into the negative RFI values to get an idea for how it behaves).
- one option would be to reflect the standard curve's behavior across the x-axis.
- perhaps equivalently, we could embrace the same underlying rationale for the logicle plot: linearize our function around 0 and extend it into the negative numbers.
  - is that the same as what you've done by assuming `p0 ~= 1` => `fl_mef = - exp(p1)*[abs(fl_rfi)**p0]`? I guess not technically, since you still have the `**p0` term. Does this introduce a discontinuitinty in the derivative of this function / make it not smooth? Keeping the `**p0` term seems like it will either introduce a discontinuity or mitigate the introduction of a discontinuity, but i'm not sure which. Would be nice to be smooth around 0, though (although I don't have an immediate application which requires that).
  - linearized version might just be: `fl_mef = - exp(p1) * abs(fl_rfi)`?
- could also include some more derivatives when you interpolate past 0? i.e. Taylor (Maclourin?) expansion? - I don't think this would be terribly useful, though
- I feel like reflecting/linearizing/assuming `p0~=1` is more appropriate than only using the real part. Invoking this complex logarithm doesn't feel appropriate for our situation; feels like a weird hack tied to this strange `p0` term in our model, not something fundamentally represented in our underlying data.

Practically speaking, though, it would help to see how these solutions look alongside each other (which again probably requires logicle plot of standard curve? can't see 0 or negative behavior on log plot. or maybe just linear plot that is zoomed in)

castillohair, line 417: Yeah, reflection of the standard curve is pretty much what I did. 

Start from

```
s(x) = exp(p[1] + p[0]*log(x))
```

Which is defined for `x > 0` only. The following

```
s2(x) = exp(p[1]) * (x**p[0])
```

Is equivalent to the first form when `x > 0`, and it is also defined for `x = 0` as `s2(x) = 0`, for `p[0] > 0` (which I think should always be the case for us). Both `s(x)` and `s2(x)` converge to zero when `x -> 0+`, so `s2(x)` is continuous from the right.

I'm gonna guess that by "reflection", you meant making an odd function (https://en.wikipedia.org/wiki/Even_and_odd_functions#Odd_functions). This would result in negative RFIs being converted into negative MEFs. I'll call this `s3(x)`. This function should satisfy `s3(x) = s2(x)` when `x >= 0`, and `s3(x) = -s2(-x)` when `x < 0`. In other words,

```
s3(x) =   exp(p[1]) * (  x **p[0]),    x >= 0
      = - exp(p[1]) * ((-x)**p[0]),    x < 0
```

From the discussion above, this is an odd function, continuous at zero with `s3(0) = s3(0+) = s3(0-) = 0`.

The definition of `s3(x)` can be combined in one single equation:

```
s3(x) = sign(x) * exp(p[1]) * (abs(x)**p[0])
```

Which is exactly what I have in the code.

castillohair, line 417: f11ad26695de896a9bac31ee5b0a507ac00624fc improves the explanation of this part.

FlowCal/mef.py
castillohair, line 477: I will not modify this since you agreed that we should keep this in log scale. But I wanted to answer some things.

> I'm also confused why there aren't negative values on the top plot. Do you only show negative numbers if necessary.

The negative interval in a logicle scale is "adaptive", in that it is changed depending on how negative the points are. This is not an axis limit effect, it affects scaling throughout all the range. For more details, check the paper.

This means that if there is no negative information, you will see that the axis starts at zero, but the rest of the plot will look pretty much like a log scale.

> I'm struggling to identify where the linear-to-log transition occurs on the scales, too.

It is a soft transition from log to linear, so most likely you won't be able to see it clearly.

FlowCal/io.py
castillohair: Changed in 6b954c408c57b9450dafc57d7004fe7126996083.

FlowCal/excel_ui.py
castillohair, line 409: Yeah, cells don't look very good with `sigma=5`.

FlowCal/plot.py
castillohair, line 1170: Another issue is that floating-point files have negative events, which will not be covered by `.range` independently of whether we start with 0 or 1. Logicle axis adapt to this, but if we manually force the lower axis limit to be equal to the lower range value, we will not show negative values and the whole point of logicle scaling is lost.

In short, the range is not very precise, and `hist_bins` does a better job at giving limits for visualization.

castillohair, line 1170: Improved comments in 96c740ee8b3365e71fa43c15a3205b38d7a91e56.

FlowCal/excel_ui.py
castillohair, line 1177: I'll take this into account when updating the documentation.

castillohair:
I'll make this brief and only show MEF'd cell plots.

Integer, linear:
s2lin_0008

Integer, log:
s2log_0008

Floating-point
s3_0008

castillohair:
It seems that everything had been addressed. I'll merge later today after some more testing.

PR #221, opened 2016-12-08
Excel UI does not require plotting directory

### LINKED ISSUE #218: Excel UI should plot to standard plot output (e.g. Jupyter notebook) if no plot directory is specified
`FlowCal.excel_ui.process_beads_table()` and `FlowCal.excel_ui.process_samples_table` should allow for `plot_dir` to be `None` (or `False`?), in which case they should render their plots to the console (or whatever the standard plotting output is).

In particular, I'm using the Python API in a Jupyter notebook, and I want to use `FlowCal.excel_ui.process_beads_table()` and have its plots rendered in the notebook, but that doesn't work in the current build. Current build fails if you specify `plot_dir=None`, which implicitly requires that you save your plots to a directory if you want to see them (default behavior if you don't specify `plot_dir` is to dump the plot files in the current directory).

In contrast, `FlowCal.mef.get_transform_fxn()` exhibits the desired behavior (if `plot==True` and `plot_dir is None`, plot without saving to file). I tested this, and indeed the plots generated by `FlowCal.mef.get_transform_fxn()` are successfully rendered to a Jupyter notebook if you specify `plot=True` and `plot_dir=None'.
* minor note: `FlowCal.mef.get_transform_fxn()` only accepts `plot_dir=None` (not `plot_dir=False`), so maybe `FlowCal.excel_ui.process_beads_table()` and `FlowCal.excel_ui.process_samples_table` should only accept `plot_dir=None` to be consistent.

JS3xton:
Closes #218.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -527,7 +527,7 @@ def process_samples_table(samples_table,
                           base_dir=".",
                           verbose=False,
                           plot=False,
-                          plot_dir="."):
+                          plot_dir=None):
     """
     Process flow cytometry samples, as specified by an input table.
 
@@ -581,7 +581,8 @@ def process_samples_table(samples_table,
         sample, and each beads sample.
     plot_dir : str, optional
         Directory relative to `base_dir` into which plots are saved. If
-        `plot` is False, this parameter is ignored.
+        `plot` is False, this parameter is ignored. If ``plot==True`` and
+        ``plot_dir is None``, plot without saving.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


JS3xton:
As of e424a3a6b0e4c27c57288f07fda2b637746467f1, all unit tests pass:

```
> python -m unittest discover
........................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 280 tests in 4.694s

OK
```

Calling `> python -m FlowCal.excel_ui -v -p` prompts the user for an input XLSX file and then correctly populates `plot_beads` and `plot_samples` folders.

Calling `> python -m FlowCal.excel_ui -v -p -i <input XLSX file>` correctly populates `plot_beads` and `plot_samples` folders.

castillohair:
I'd like to see test results showing that plots are actually rendered in the standard output when `plot_dir=None`. If you have a python notebook that you can upload so I can reproduce the test, it would be even better.

Code looks fine otherwise.

JS3xton:
See images below and attached (zipped) Jupyter notebook.

console_plotting_1

console_plotting_2

[FlowCal_excel_ui_console_plotting.zip](https://github.com/taborlab/FlowCal/files/681712/FlowCal_excel_ui_console_plotting.zip)

Issue #222, opened 2016-12-10
Add kwargs catch-all to Excel UI process_beads_table() which gets passed to mef.get_transform_fxn()


JS3xton:
I'm using the Excel UI via the Python API to process bead files, and it's excluding some bead peaks that I don't want excluded in the subpopulation selection step. To get around this, I wanted to turn off the automated subpopulation selection step and just manually specify which peaks to keep and which peaks to exclude by placing `None` in the comma-separated list of MEF values on the Beads tab in the XLSX sheet. To achieve this, though, I had to copy the entire `excel_ui.process_beads_table()` function source code into my context (Jupyter notebook), update module names to match my namespace, and then add in 1 line to set `mef.get_transform_fxn(...,selection_fxn=None)`.

I think a better solution would be to add a `kwargs` catch-all parameter to `excel_ui.process_beads_table()` which would be passed to the internal `mef.get_transform_fxn()` call. This offers up the much simpler and generally much more flexible solution to my problem stated above:

```python
beads_samples, mef_transform_fxns, mef_outputs = excel_ui.process_beads_table(
    beads_table,
    instruments_table,
    get_transform_fxn_kwargs={'selection_fxn':None}    # <---
    base_dir=input_dir,
    verbose=verbose,
    plot=plot,
    plot_dir='plot_beads',
    full_output=True)
```

castillohair:
Merged into branch develop.

PR #223, opened 2016-12-11
Added kwargs to Excel UI process beads table.

JS3xton:
See #222.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -212,12 +212,12 @@ def write_workbook(filename, table_list, column_width=None):
 
 def process_beads_table(beads_table,
                         instruments_table,
-                        get_transform_fxn_kwargs={},
                         base_dir=".",
                         verbose=False,
                         plot=False,
                         plot_dir=".",
-                        full_output=False):
+                        full_output=False,
+                        get_transform_fxn_kwargs={}):
     """
     Process calibration bead samples, as specified by an input table.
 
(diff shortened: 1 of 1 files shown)


JS3xton:
As of e1111cf, all unit tests pass:

```
> python -m unittest discover
........................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 280 tests in 5.213s

OK
```

Tested manually and was able to turn off subpopulation selection step:
```python
FlowCal.excel_ui.process_beads_table(beads_table, instrument_table, plot=True)
```
yielded:
populations_fl3_b1

```python
FlowCal.excel_ui.process_beads_table(beads_table, instrument_table, plot=True, get_transform_fxn_kwargs={'selection_fxn':None})
```
yielded:
populations_fl3_b1

(FL3 channel; top peak was being explicitly excluded in the input XLSX sheet)

castillohair:
Just to confirm, there are two peaks to the right being excluded in the second plot of your example. Are you explicitly excluding both?

JS3xton:
Yep. This should be the `Beads` tab of the XLSX sheet for that plot:

excel_ui

FlowCal/excel_ui.py
```diff
@@ -212,6 +212,7 @@ def write_workbook(filename, table_list, column_width=None):
 
 def process_beads_table(beads_table,
                         instruments_table,
+                        get_transform_fxn_kwargs={},
```
castillohair: I think this should be the last parameter. This is such an edge case that I don't think changing the order of parameters is appropriate.

castillohair (changes requested):
The order of parameters is the only complain I have.

JS3xton (commented):
updated as per your request

Issue #236, opened 2017-02-26
Cryptic error message as a result of bad gating fraction


castillohair:
When using the Excel UI, using an invalid gating fraction (e.g. 1.5) gives the following cryptic error message:

```
Sample ID S0002...
Loading file "FCFiles/data002.fcs"...
Performing data transformation...
Performing gating...
Traceback (most recent call last):
  File "C:\Users\casti\Anaconda2\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Users\casti\Anaconda2\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 1468, in <module>
    plot=args.plot)
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 1400, in run
    plot_dir='plot_samples')
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 791, in process_samples_table
    full_output=True)
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\gate.py", line 487, in density2d
    Nidx = np.nonzero(csvH >= n)[0][0]    # we want to include this index
IndexError: index 0 is out of bounds for axis 0 with size 0
```

The error says nothing about what the problem is, and it crashes the execution of the whole program. Instead, a more descriptive error message should be printed in the "Analysis notes" column.

castillohair:
Addressed in 81a6a9756cb71b8c1818dd4f7392deadf291e00d.

castillohair:
Solution merged into develop, see #237.

PR #237, opened 2017-02-26
Added proper handling of improper gating fraction values

castillohair:


JS3xton:
My initial thoughts:
- I agree that the current behavior is not clear and am in favor of clarifying it.
- why is 0 not included in the acceptable interval? This is not really a huge deal, but I might reasonable expect simply an empty output (e.g. empty numpy array or FCSData object from the gate module functions? And maybe downstream failure in the Excel UI because I think we arbitrarily required 250 events at some point in the Excel UI) instead of an error if `gate_fraction` is 0.
- I think this parameter check should be covered in a unit test. Specifically, I would expect both a unit test on the gate module function as well as a separate unit test on the Excel UI.
- the "before" behavior is an error message, which is described in #236. Would be nice to see the "after" behavior (e.g. a screenshot of the Excel UI with the "Analysis notes" column populated).

JS3xton:
I was also originally gonna say we should explicitly unit test the edges cases of `gate_fraction=0.0` and `gate_fraction=1.0`, but looks like I already put that in a while back: [test_gate.py#L833](https://github.com/taborlab/FlowCal/blob/81a6a9756cb71b8c1818dd4f7392deadf291e00d/test/test_gate.py#L833).

This again calls into question what the behavior should be with `gate_fraction=0.0`, though. It looks like the current pull request code fails those 3 `gate_fraction=0.0` unit test.

castillohair:
- My reasoning for not including 0 is that if you want zero events, you don't really need to use the function. You just make an empty array. I don't really mind allowing it, though.
- I can make a unit test for `gate.density2d()`. Error conditions in the Excel UI do not have unit tests (this will probably require a new set of erroneous Excel files), so I'll pass on it. But more thoroughly testing of the Excel UI errors could be done in another issue/pull request.
- I'll upload a screenshot of the Excel UI and the new error message soon.
- I guess if I allow `gate_fraction=0.0` the unit tests will be successful? I could also delete those tests. Again, I don't really care either way.

JS3xton:
> - My reasoning for not including 0 is that if you want zero events, you don't really need to use the function. You just make an empty array. I don't really mind allowing it, though.
> - I guess if I allow `gate_fraction=0.0` the unit tests will be successful? I could also delete those tests. Again, I don't really care either way.

- I think there should be a unit test of the `gate_fraction=0.0` edge case regardless of whatever behavior we settle on, so I'm not a fan of deleting the unit tests outright.
- I tend to favor allowing `gate_fraction=0.0` (I would simply expect empty data structures, not an error, and I think the function should behave as expected, even with uncommon or arguably unnecessary inputs), but my preference is not strong.

> - I can make a unit test for `gate.density2d()`. Error conditions in the Excel UI do not have unit tests (this will probably require a new set of erroneous Excel files), so I'll pass on it. But more thoroughly testing of the Excel UI errors could be done in another issue/pull request.

Sounds good. Just adding a `gate` module unit test seems good enough to me.

> - I'll upload a screenshot of the Excel UI and the new error message soon.

👍

castillohair:
- I modified `gate.density` to allow for `gate_fraction=0`.
- I added 2 unit tests for confirming that the error is properly thrown.
- Here is the screenshot of the error message in the Excel UI:

gate-fraction-error

I'll merge tomorrow night if there are no more comments.

Issue #245, opened 2017-04-05
Density gating should be optional in the Excel UI


castillohair:
I think an additional column called "Scatter Gating" should be added, in which the option "Density" would result in the density gate that we currently use. "None" would result in no FSC/SSC gating. Future methods could include "Gaussian", similar to what TASBE does.

If the column "Scatter gating" is not present, it would be cleaner to not use any gating. But to preserve backwards compatibility we could default to density gating. I'm fine either way.

PR #246, opened 2017-04-05
Implemented optional density gating in the Excel UI.

castillohair:
Functionality is implemented as described in #245. Some parts of the documentation need to be updated once we agree on the functionality and the actual code.

In summary, both sheets "Samples" and "Beads" now support an additional column called "Scatter Gate". A value of "Density" results in density gating, and "None" or "" result in no gating. The absence of the column "Scatter Gate" results in density gating, so we don't break the current behavior. This only controls density gating, and high/low and start/end gating are still applied.

With current branch (018d457fd55624722e49aa815242b18ba192c5f2) the attached input excel [input excel file](https://github.com/taborlab/FlowCal/files/899869/experiment.xlsx) results in the attached
[output file](https://github.com/taborlab/FlowCal/files/899870/experiment_output.xlsx).

This is one of the plots when density gate is not applied:

s0002

Notice the lack of a contour in FSC/SSC. Start/end and high/low gating are still applied, therefore the histogram still shows gated/ungated populations. In this case, these are barely different.


FlowCal/excel_ui.py
```diff
@@ -379,15 +379,45 @@ def process_beads_table(beads_table,
...
+                    gate_method = ''
+                else:
+                    gate_method = str(beads_row['Scatter Gate'])
+
+                if gate_method.lower().strip() == 'density':
+                    # Density gate by default
+                    beads_sample_gated, __, gate_contour = FlowCal.gate.density2d(
+                        data=beads_sample_gated,
+                        channels=sc_channels,
+                        gate_fraction=beads_row['Gate Fraction'],
```
JS3xton: Seems like you might want to be more robust to the presence / absence of a Gate Fraction column (I'm not actually sure how this code behaves if the Gate Fraction column is omitted).

Also begs the larger question of how should gate parameters be handled in the general case?

JS3xton:
I generally like the idea. My thoughts are kinda all over the place on the specifics, though, but I'll try and distill them:
- I like the idea of granting greater resolution/control in the Excel UI over the pre-processing/gating pipeline. I think we could possibly go further than what currently exists, but doing so elegantly, succinctly, and in a backwards compatible manner may be nontrivial. Possible extensions of current implementation:
  - Other gating strategies? A Threshold gate comes to mind as simple but particularly useful (although it could be argued that the Threshold gate is a subset of the more general `high_low` gate).
  - Might be nice to also have a "Fluorescence Gating" column (there have been times when I've wanted to employ a low-threshold fluorescence gate to get rid of noise that was not removed via density gating on some very bright cell events). Can Fluorescence Gating be combined with Scatter Gating, though? In what order are the gates applied? (my initial expectation is that all Scatter Gates should be applied first and then all Fluorescence Gates. This may only matter for more complex gating strategies that are informed by the data, though, e.g. Density and Gaussian, not something simple like a Threshold gate).
  - Could potentially be very powerful if you could compose gates (e.g. comma- or semicolon-separated list of recognized gates?). I think you could pull a lot of users up out of the Python API if you could get that functionality exposed via the Excel UI.
  - How do you handle gate parameters? I'm envisioing semicolon-delimited list for each gate (if multiple gates are allowed) and within each semicolon list element a comma-delimited list of gate parameters which are forwarded to the gate function. Does the comma-delimited list have to rely on argument position? Or could you pass key-value pairs in the comma-delimited list? (e.g. - "Scatter Gate" column="High Low, High Density", "Scatter Gate Parameters" column="high=500,low=200;gate_fraction=0.85", etc.). The order of the gates would also be important (Density before vs. after an aggressive `high_low` gate could result in very different things). Could even have multiple instances of the same gate (e.g. "Density, High Low, Density". Not actually sure why I would want to do that, but would be very powerful if the Excel UI could tolerate that input).
  - Might be kinda useful to accept an arbitrary "user-defined" gate function. Not sure about the implementation details, but maybe something like a Python file (module) which the user has to ensure is in the same directory as the Excel UI XLSX file and they put the name of the module or function in the "Scatter Gate" or "Fluorescence Gate" column and FlowCal attempts to find that user-defined gating function.
- The purist in me doesn't like differentiating between density gating and `high_low`/`start_end` gating. They all remove events (although not all of them operate on the scatter channels). I acknowledge that `high_low` and `start_end` are kinda lower significance gates and arguably appropriate in most situations, but I like the idea of a more flexible input facilitated by this "Scatter Gate" column (see discussion above). I've also had situations where I did *not* want to do `high_low` gating (quantifying 2 different populations where one of my populations was saturating. FSCvSSC gate made me confident I was isolating cells, so I did *not* want to throw away events that saturated low or high in my fluorescence channel).
- I don't love the default behavior being to density gate, but I appreciate that we should keep it backwards compatible and consistent with the behavior described in the paper. Perhaps it's fine; the "Scatter Gate" column can be thought of as an Advanced Setting, and on the Basic Setting it should try and do everything for you (i.e. density gate).
 - Might be better to name the gating strategies more descriptively. "Density" doesn't tell you if it's high density or low density events which are being selected for. Maybe "High Density" would be better? Thinking about the Threshold gate, too, perhaps "Above" and "Below" gates would be more clear?

castillohair:
Thanks for the feedback. I'll try to reply to everything:
> Other gating strategies? A Threshold gate comes to mind as simple but particularly useful (although it could be argued that the Threshold gate is a subset of the more general high_low gate).

Threshold is a good idea, since it's something that old-school flow cytometry users would recognize easily. It could be called Threshold, and be implemented under the hood with `high_low`. We could also have an explicit "High/Low" gating method, which allows the user to specify two methods.

> Might be nice to also have a "Fluorescence Gating" column (there have been times when I've wanted to employ a low-threshold fluorescence gate to get rid of noise that was not removed via density gating on some very bright cell events). Can Fluorescence Gating be combined with Scatter Gating, though? In what order are the gates applied?

I think this is a good idea, and it would help our lab people doing sensing in soil/fecal samples. I would agree with having the scatter gate applied before the fluorescence gate. I'm not sure if this should be in this issue/pull request. I guess we'll see how the code develops, and if it's straightforward we can do it here. Otherwise, we can open a new issue.

> Could potentially be very powerful if you could compose gates (e.g. comma- or semicolon-separated list of recognized gates?). I think you could pull a lot of users up out of the Python API if you could get that functionality exposed via the Excel UI.

This might get too complicated, in my opinion. But could be doable.

> How do you handle gate parameters? I'm envisioing semicolon-delimited list for each gate (if multiple gates are allowed) and within each semicolon list element a comma-delimited list of gate parameters which are forwarded to the gate function. Does the comma-delimited list have to rely on argument position? Or could you pass key-value pairs in the comma-delimited list? (e.g. - "Scatter Gate" column="High Low, High Density", "Scatter Gate Parameters" column="high=500,low=200;gate_fraction=0.85", etc.)

I don't like this because:
1) This will be problematic if you have specified two or more gating methods with parameters that have the same name.
2) This eliminates the Excel feature of making series and running calculations on cell values, since the cell value is now a long string.

My proposal would be to have additional columns for each parameter. If we allow several gates, an example would be "Scatter Gate": "High Low, High Density", "Scatter Gate 1: High": 1e5, "Scatter Gate 1: Low": 1e3, "Scatter Gate 2: Gate Fraction": 0.5.

> Could even have multiple instances of the same gate (e.g. "Density, High Low, Density". Not actually sure why I would want to do that, but would be very powerful if the Excel UI could tolerate that input).

I think my proposal above would naturally accept this kind of behavior.

> Might be kinda useful to accept an arbitrary "user-defined" gate function. Not sure about the implementation details, but maybe something like a Python file (module) which the user has to ensure is in the same directory as the Excel UI XLSX file and they put the name of the module or function in the "Scatter Gate" or "Fluorescence Gate" column and FlowCal attempts to find that user-defined gating function.

I think this could be doable with a straighforward extension of my proposal above.

> The purist in me doesn't like differentiating between density gating and high_low/start_end gating. They all remove events (although not all of them operate on the scatter channels). I acknowledge that high_low and start_end are kinda lower significance gates and arguably appropriate in most situations, but I like the idea of a more flexible input facilitated by this "Scatter Gate" column (see discussion above). I've also had situations where I did not want to do high_low gating (quantifying 2 different populations where one of my populations was saturating. FSCvSSC gate made me confident I was isolating cells, so I did not want to throw away events that saturated low or high in my fluorescence channel).

Remember that high/low is only applied as preprocessing in files with integer data types, because the saturated events do not contain correct values. I think this goes beyond optional data processing and should never be skipped. That said, I would be ok with exposing an option to disable this, not via the "Scatter gate" column but via an additional column (e.g. "Remove Saturating Events": TRUE or FALSE)

> Might be better to name the gating strategies more descriptively. "Density" doesn't tell you if it's high density or low density events which are being selected for. Maybe "High Density" would be better? Thinking about the Threshold gate, too, perhaps "Above" and "Below" gates would be more clear?

I'm not too concerned about the names, as long as the methods are well described in the documentation. I would favor keeping the name "Density", since it's what we used in the paper. "Threshold", I think, it's easily recognizable by old-school flow cytometer users, but I'm more flexible regarding names of things we haven't implemented yet.

castillohair:
Just realized that contours are gonna be really hard to draw if we allow multiple gates. We might want to stick to only one to not make our lives too hard.

JS3xton:
>1. This will be problematic if you have specified two or more gating methods with parameters that have the same name.
>2. This eliminates the Excel feature of making series and running calculations on cell values, since the cell value is now a long string.

I don't understand either of these.

(1) With multiple gating methods, the collection of parameters is still positional, so even if there are methods that employ parameters of the same name, those parameters should still be uniquely routed to the correct gating method based in their position in the "outer" delimited list (e.g. the semicolon-delimited list). So I don't think I understand your concern. An example might help me. The code for this would really just be a little parsing (split on `;` and on `,`) and then passing the results to appropriate functions. And if you have multiple gating methods that all use the same parameter (e.g. `gate_fraction`), I still don't understand how that's a problem.

(2) I don't understand what you mean by this.

> My proposal would be to have additional columns for each parameter.

This feels less elegant to me / kinda kludge, but I think it would work. The number and names of the columns would change arbitrarily, whereas encapsulating all gating specifications within 2 columns kinda standardizes the Excel file columns. We've already embraced "non-standardized" Excel columns, though, with the specification of MEF Units and MEF Values, so additional columns is actually arguably more consistent with the rest of the Excel UI.

> because the saturated events do not contain correct values. I think this goes beyond optional data processing and should never be skipped.

The values are not incorrect, they're just saturated. That's still potentially very useful information. See use case previously described.

Furthermore, I actually kinda feel unjustified in throwing out saturated events now-a-days. (1) You should do a good job of isolating cell events, which should get rid of most saturating events (if your detector ranges are appropriately tuned). (2) If you do a good job of isolating cell events but you still have saturating events, then you actually have cells with fluorescence that cannot be detected in the current instrument detection range on the channel you care about. I think the most appropriate way to handle that should be left to the user and should not be done automatically (@thoreusc's pruning algorithm comes to mind). I think the user *should* be able to discard those events if they deem that appropriate, but they should have to do it explicitly (e.g. via a Fluorescence Gating Threshold gate).

> I would be ok with exposing an option to disable this, not via the "Scatter gate" column but via an additional column (e.g. "Remove Saturating Events": TRUE or FALSE)

I prefer a more natural mechanism that is enabled by a flexible gate specification scheme, so I'm not a huge fan of an extra explicit column for this. This column would also be extraneous for files with floating point data types?

I think having a mechanism for specifying whether or not saturating events should be removed is better than not having a mechanism, though.

> Just realized that contours are gonna be really hard to draw if we allow multiple gates. We might want to stick to only one to not make our lives too hard.

I don't understand why. `gate.density2d()` already returns a list of lines representing the gate contour. I think Threshold gate simply needs to also define it's own 2D numpy array representation of itself, and then you can just concatenate and illustrate all of the 2D numpy array gate representations.

castillohair:
This has been some useful discussion. I've made a couple of decisions on how to go about this, which I will summarize below:
1. The original intention of this request was to be able to skip using density gate from the Excel UI. In order to ever get something done, this pull request will be considered complete when this goal is achieved and documentation is updated to reflect this.
2. I also wanted to do this in a way that would allow specifying other "scatter gates" as we incorporate them into FlowCal (e.g. Gaussian, or issue #267 which I think it's an interesting idea). I think it is straighforward to implement item 1 in a way that facilitates specifying gating algorithms once the functions are implemented in the `gate` module. However, implementation of each additional gating algorithm should go through its own issue/pull request (I will open an issue for the Threshold gate).
3. An additional interesting idea was to have the ability to specify multiple gates. While I think this is a good concept, there is a detail that we haven't resolved to my satisfaction: the graphical representation (see below for more details). In addition, we don't have anything but density gate in FlowCal right now, so allowing for multiple gates might not currently be that useful. I think, however, that once item 1 is implemented, the expansion to allow for multiple gates would not be difficult. This also requires that the new Excel file format be well thought to anticipate this, and this discussion has been excellent for doing that. I think the best thing to do here would be to open a new issue for "multiple scatter gates from the excel file", and continue the discussion there.
4. Another idea was to allow for everything above, but with fluorescence channels. Again, I think this builds from all the ideas above, and would be a good thing to keep discussing in a separate issue.
5. Finally, you also mentioned that you would want to be able to disable discarding the saturating events. For the sake of getting something done out of this issue, I think this should also be separated into its own issue.

I will respond to some of the comments from your last response:

> >1. This will be problematic if you have specified two or more gating methods with parameters that have the same name.
> >2. This eliminates the Excel feature of making series and running calculations on cell values, since the cell value is now a long string.

> I don't understand either of these.

> (1) With multiple gating methods, the collection of parameters is still positional, so even if there are methods that employ parameters of the same name, those parameters should still be uniquely routed to the correct gating method based in their position in the "outer" delimited list (e.g. the semicolon-delimited list). So I don't think I understand your concern. An example might help me. The code for this would really just be a little parsing (split on `;` and on `,`) and then passing the results to appropriate functions. And if you have multiple gating methods that all use the same parameter (e.g. `gate_fraction`), I still don't understand how that's a problem.

Sorry, I didn't catch that it was supposed to be a list of lists. I guess it makes more sense now.

> (2) I don't understand what you mean by this.

I meant that things are nicer in Excel when you have a single number on a single cell, because you can operate on them from other cells or use autofilling. Having long lists of lists as cell values makes these things more difficult, if not impossible.

> > My proposal would be to have additional columns for each parameter.

> This feels less elegant to me / kinda kludge, but I think it would work. The number and names of the columns would change arbitrarily, whereas encapsulating all gating specifications within 2 columns kinda standardizes the Excel file columns. We've already embraced "non-standardized" Excel columns, though, with the specification of MEF Units and MEF Values, so additional columns is actually arguably more consistent with the rest of the Excel UI.

I guess the feeling of "kludginess" is subjective in this case, because having a list of lists feels very inelegant to me. As you said, we already use columns with non-fixed names, so using individual columns per parameter would be more consistent.

>> because the saturated events do not contain correct values. I think this goes beyond optional data processing and should never be skipped.

>The values are not incorrect, they're just saturated. That's still potentially very useful information. See use case previously described.

I guess this goes into semantics, but saturated events are, almost by definition, numerically incorrect. Given that the main outputs of the Excel UI are statistics, these are going to be incorrect when saturated events are used.

> I think the most appropriate way to handle that should be left to the user and should not be done automatically (@thoreusc's pruning algorithm comes to mind).

The user can still skip this step and implement more complicated algorithms with the python UI.

>> I would be ok with exposing an option to disable this, not via the "Scatter gate" column but via an additional column (e.g. "Remove Saturating Events": TRUE or FALSE)

> I prefer a more natural mechanism that is enabled by a flexible gate specification scheme, so I'm not a huge fan of an extra explicit column for this. This column would also be extraneous for files with floating point data types?

> I think having a mechanism for specifying whether or not saturating events should be removed is better than not having a mechanism, though.

I think it is a good compromise to add an additional column to allow to enable or disable saturating events. I maintain that it should not be part of the scatter gating specification in principle for the reasons specified above. But a more practical reason is that the removal of saturating events is currently performed in every relevant fluroescence channel as well. Should we wait until we are capable of specifying fluorescence gates from the Excel UI? I think it is more practical to just have a single flag.

> > Just realized that contours are gonna be really hard to draw if we allow multiple gates. We might want to stick to only one to not make our lives too hard.

> I don't understand why. `gate.density2d()` already returns a list of lines representing the gate contour. I think Threshold gate simply needs to also define it's own 2D numpy array representation of itself, and then you can just concatenate and illustrate all of the 2D numpy array gate representations.

I think this can result in a graph that is very hard to interpret, especially as you increase the number of gates. Maybe a good way to do this would be to draw each individual contour in grey, and a contour that only delineates the final selected population with black. How to calculate the final contour is something I don't know. Another idea would be to make the events outside the final gated region more transparent, and then you can draw as many gating contours as you want, so that the intersection of these contours is very clear. I think this would need some testing, and it would probably be wise to wait for FlowCal to have more gates. This is why I'd rather open another issue for this.

Finally, we discussed in person how to go about supporting the current Excel file format (density gate by default, Gate Fraction as a mandatory column). We thought that a flag specified when calling the Excel UI would be the cleanest approach, although we weren't sure about the specifics. I will use a flag called `--auto-density-gate` that makes the Excel UI work as it does now, unless there are some objections.

To finish, I will modify the code submitted to this request to comply with items 1 and 2, as specified above. I will open new issues for Threshold gating, multiple scatter gates, fluorescence gates, and the ability to disable automatic discarding of saturated events.

PR #253, opened 2017-05-05
Better support for supplemental TEXT segments.

### LINKED ISSUE #247: Supplemental TEXT segment should not be required to start with delimiter
## Issue Summary
FlowCal fails to create an `FCSData` object (or an `FCSFile` object) for an FCS file which contains a supplemental TEXT segment which does *not* start with the `delimeter` character used in the primary TEXT segment.

## Error Message
```python
d = fc.io.FCSData('stext.fcs')

  File "build\bdist.win-amd64\egg\FlowCal\io.py", line 1425, in __new__
    
  File "build\bdist.win-amd64\egg\FlowCal\io.py", line 636, in __init__
    
  File "build\bdist.win-amd64\egg\FlowCal\io.py", line 176, in read_fcs_text_segment
  
ValueError: segment should start with delimiter
```
## FCS File Contents
Beginning of primary TEXT segment:

`|$BEGINANALYSIS|0000000000000000|$ENDANALYSIS|0000000000000000|$BEGINDATA|0000000000004161|$ENDDATA|0000000004204160|$BEGINSTEXT|0000000004204161|$ENDSTEXT|0000000004292439|$NEXTDATA|0000000000000000|$DATATYPE|I|$MODE|L|$PAR|21|$BYTEORD|1,2,3,4|`...

(delimiter = `|`)

Beginning of supplemental TEXT segment:

`SORTSTATS|AAEAAAD/////AQAAAAAAAAAMAgAAAEZMaWJDeXRvUHJvdG9jb2wsIFZlcnNpb`...

## Notes on Supplemental TEXT Segment and Delimiter
* supplemental TEXT segment introduced in [FCS3.0](http://murphylab.web.cmu.edu/FCSAPI/FCS3.html).
    > 2.2.3 ... New to FCS3.0 is the inclusion of an optional supplemental TEXT segment.
* documentation on delimiter from [FCS3.0](http://murphylab.web.cmu.edu/FCSAPI/FCS3.html):
    > 2.2.15 The "delimiter" is the first character of the primary TEXT segment and is subsequently placed in the primary TEXT, supplemental TEXT and ANALYSIS segments to separate keywords from keyword values. The delimiter can be any ASCII character.

    > 3.2.5 The first character in the primary TEXT segment is the ASCII delimiter character. This character must also be used as the delimiter in the ANALYSIS and supplemental TEXT segments.

    > 3.2.6 The delimiter is placed at the start and end of a keyword value.

    > 3.2.7 The delimiter may not be the first character in a keyword or keyword value. If the delimiter appears in a keyword or keyword value, it must be immediately followed by a second delimiter. For example, "$SYS/RSX-11//M/" shows a value of RSX-11/M for the keyword $SYS. Since null (zero length) keywords or keyword values are not permitted, two consecutive delimiters can never occur between a value and a keyword.
* changes to delimiter in [FCS3.1](http://isac-net.org/PDFS/90/9090600d-19be-460d-83fc-f8a8b004e0f9.pdf):
    > The delimiter character has been restricted to any single-byte ASCII character from the
range of 1-126 (01-7E hex).

## My Thoughts
I originally interpreted `3.2.6` to mean that you needed the delimiter on both sides of both `keys` and `values` (e.g. - `/k/v/`, which requires 3 `/` delimiter characters). Now, I think the intention is to specify that the delimiter is only required to be on both sides of the `values`? (e.g. `k/v/`) This really only matters for the first ~~and last~~ keyword-value pair~~s~~. This new interpretation of `3.2.6` along with `2.2.15` and `3.2.5` suggest that the supplemental TEXT can start with a `key` (instead of the `delimiter`, as previously assumed).

JS3xton:
## Summary
- Fixes #247 (supplemental TEXT segments are no longer required to start with delimiter).
- Also fixes related bug which arises when a TEXT segment delimiter character occurs in a TEXT segment keyword or keyword value.
- Test FCS files which were previously parsed without error now fail to parse, possibly necessitating a new, more flexible/lenient approach to parsing FCS files?
## Changes to Keyword/Value Delimiter Reconstruction Algorithm
[Old Implementation](https://github.com/taborlab/FlowCal/blob/4420e15428d13923e8da32f85ae7676adfeade81/FlowCal/io.py#L186) (`develop`, 4420e15428d13923e8da32f85ae7676adfeade81): a sequence of consecutive delimiter characters must be escaped once (as opposed to escaping each delimiter character individually).
- e.g. - the Python dictionary `{'key///1':'value1'}` maps to the following primary TEXT segment: `'/key////1/value1/'`

[New Implementation](https://github.com/taborlab/FlowCal/blob/6fcce1c24adadc2110ab782f9063f6649a7c3dbd/FlowCal/io.py#L212) (6fcce1c24adadc2110ab782f9063f6649a7c3dbd): each instance of a delimiter character that exists in a keyword or keyword value must be escaped individually.
- e.g. - the Python dictionary `{'key///1':'value1'}` maps to the following primary TEXT segment: `'/key//////1/value1/'`

The FCS standards are frustratingly unclear on this distinction. I spoke with @castillohair, though, and we both agree that the New Implementation is more appropriate.

Note: the issue of delimiter characters in keywords or keyword values was first addressed in #203 and #206.

## Validation
To validate these bug fixes, I wrote [77 unit tests](https://github.com/taborlab/FlowCal/blob/6fcce1c24adadc2110ab782f9063f6649a7c3dbd/test/test_io.py#L108).
- The primary TEXT segment tests run against `develop` (4420e15428d13923e8da32f85ae7676adfeade81) (26 tests total) result in: `FAILED (failures=9, errors=7)`
- The primary and supplemental TEXT segment tests (77 tests total) all pass against 6fcce1c24adadc2110ab782f9063f6649a7c3dbd.

In addition to unit tests, I also wrote a script which attempts to load FCS files from the following FCS file repository: https://flowrepository.org/id/FR-FCM-ZZZ4. I also included some possible troublemaker files from a flow cytometry interlab study currently under way (`N007.fcs`, `N009.fcs`, `N016.fcs`, `N022.fcs`).

Script (`io_test.py`, cannibalized from #203 script):
```python
import FlowCal as fc
import os
import numpy as np

flow_dir = 'fcs_files'

files = os.listdir(flow_dir)

for idx,filename in enumerate(files):
    print('\n{0:02d} : {1:}'.format(idx,filename))
    try:
        f = fc.io.FCSFile(os.path.join(flow_dir,filename))

        # Version
        print('-version:\t\t\t\t\t{}'.format(f.header.version))

        # Number of channels and datatype
        num_chan = int(f.text['$PAR'])
        print('-num channels ($PAR):\t\t{}'.format(num_chan))
        print('-$DATATYPE:\t\t\t\t\t{}'.format(f.text['$DATATYPE']))

        # Look at bit widths
        param_bit_widths = [int(f.text['$P{0}B'.format(p)])
                            for p in xrange(1,num_chan+1)]
        unique_bw = np.unique(param_bit_widths)
        print('-unique bit widths ($PnB):\t({})'.format(','.join(str(u) for u in unique_bw)))

        # Supplemental TEXT segment
        stext_begin = f.text.get('$BEGINSTEXT')
        stext_end = f.text.get('$ENDSTEXT')
        print('-supplemental TEXT?:\t\t($BEGINSTEXT={},$ENDSTEXT={})'.format(stext_begin, stext_end))

        # ANALYSIS segment
        print('-ANALYSIS segment?:\t\t\t{}'.format(len(f.analysis)>0))

    except Exception as e:
        print('E:{}'.format(e))
```

Diff of script outputs run under `develop` (4420e15428d13923e8da32f85ae7676adfeade81) and 6fcce1c24adadc2110ab782f9063f6649a7c3dbd:

6fcce1c_4420e15_diff

### Analysis
Summary of diff results: three different Accuri files (`02`, `03`, `04`) now fail to parse, the Applied Biosystems file (`06`) fails with a different error message, and the problematic interlab study file (`N016.fcs`) (which ultimately brought this issue to light) is now parsed where it previously threw an error.

Taking a closer look at the 3 Accuri files that fail, they all seem to end their primary TEXT segments in a prohibited fashion:
 - `02` primary TEXT segment ending: `'.../#P14MaxUsefulDataChannel/1279//'`
 - `03` primary TEXT segment ending: `'.../#P14MaxUsefulDataChannel/1567//'`
 - `04` primary TEXT segment ending: `'.../#P14MaxUsefulDataChannel/4319//'`

Note the double `/` at the end (which happens to be tested in [this unit test](https://github.com/taborlab/FlowCal/blob/6fcce1c24adadc2110ab782f9063f6649a7c3dbd/test/test_io.py#L390)). According to the FCS standards:
> Since null (zero length) keywords or keyword values are not permitted, two consecutive delimiters can never occur between a value and a keyword.

Furthermore, the 4th Accuri file does *not* suffer the double `/` character terminating its primary TEXT segment and is parsed successfully:
- `05` primary TEXT segment ending: `'.../$ENDANALYSIS/0/#SPACERS/000/'`

The Applied Biosystems file (`06`) appears to have empty values in its primary TEXT segment, which is prohibited. This TEXT segment could almost still be parsed incorrectly (joining keywords with a delimiter character when the value is empty) if it weren't for the fact that the *last* value is empty (I didn't count all of the keyword-value pairs, so it's possible that it would still fail due to an odd number of keyword-value pairs after keywords with empty values were erroneously join via delimiters).

According to my best interpretation of a strict reading of the FCS standards, the file behaviors described above are correct and these bug fixes are consistent with the standards.

## Proposal: Warnings for Recoverable Parsing Failures?

While being technically correct is nice, it can sometimes be unnecessarily limiting. In the context of this bug fix, it may very well be possible to recover an ill-formed TEXT segment if some of the FCS standard constraints are relaxed. If such flexibility is embraced, I think the `FCSFile` object should throw a `warning` when it is forced to recover from a parsing failure. I'm not sure what form such flexibility would take in practice, though (multiple `read_fcs_text_segment()` functions? A single, more sophisticated `read_fcs_text_segment()` which makes multiple passes on ill-formed TEXT segments?), and supporting these edge cases may require a significant amount of programming effort which may not be justified given current usage patterns.

I will say that I think this approach of raising warnings upon deviation from the FCS standards would also be useful for other segment parsing functions (e.g. `read_fcs_data_segment()` with those Accuri files again, which necessitated [this hack](https://github.com/taborlab/FlowCal/pull/206/commits/c5f1b9a5404d9578daf3a8f20dfe55ed93710115)).

### DIFF
--- a/FlowCal/io.py
+++ b/FlowCal/io.py
@@ -114,8 +114,8 @@ def read_fcs_text_segment(buf, begin, end, delim=None):
     -------
     text : dict
         Dictionary of key-value entries extracted from TEXT segment.
-    delim : str
-        String containing delimiter character.
+    delim : str or None
+        String containing delimiter or None if TEXT segment is empty.
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


JS3xton:
Implementation rationale that is relevant to those Accuri files:

https://github.com/taborlab/FlowCal/blob/6fcce1c24adadc2110ab782f9063f6649a7c3dbd/FlowCal/io.py#L316

Again, the best ruling that I can get on the legality of that `/k1/v1//` pattern is:
>3.2.7 The delimiter may not be the first character in a keyword or keyword value. If the delimiter appears in a keyword or keyword value, it must be immediately followed by a second delimiter. For example, "$SYS/RSX-11//M/" shows a value of RSX-11/M for the keyword $SYS. **Since null (zero length) keywords or keyword values are not permitted, two consecutive delimiters can never occur between a value and a keyword.**

JS3xton:
Notes I took in a TXT file while coding this pull request.

[supp_text_notes.txt](https://github.com/taborlab/FlowCal/files/978337/supp_text_notes.txt)

JS3xton:
I relaxed the FCS standard constraints on ill-formed TEXT segments which have an extra delimiter at the end of the segment in b8350213f5a1b323e6be2a878c99d6cda9ed671b and reran `io_test.py`.

Diff of script outputs run under develop (4420e15) and b8350213f5a1b323e6be2a878c99d6cda9ed671b:

b835021_4420e15_diff

Now, the only change is that the problematic interlab study file (`N016.fcs`) is parsed successfully. A new warning was thrown (which was sent to `stderr` and doesn't show up in the diff), but only *one* warning was thrown instead of 3 for the Accuri files and 1 for the Applied Biosystems - Attune.fcs file, even though all 4 files are invoking that use case (which I checked). I'm assuming this is due to some wrinkle about warnings that I don't understand.

I currently left a `warning` for ill-formed TEXT segments with that end with 2 delimiters, but I'll leave it up to @castillohair to decide how best to inform the user of recovery from parsing failure.

castillohair:
I'm ok with using a `warning`.

I verified that this works as well as before with files from https://flowrepository.org/id/FR-FCM-ZZZ4. I will now merge.

Issue #239, opened 2017-03-13
more explicit error message when user tries to enter different number of MEF values into get_transform_fxn()


JS3xton:
see #146

PR #254, opened 2017-05-07
More explicit error message in Excel UI.

JS3xton:
Added more explicit error message if the user doesn't specify the same number of MEF Values for all channels of a bead entry in the Beads tab of the Excel UI.

See #239.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -480,7 +480,8 @@ def process_beads_table(beads_table,
                     raise ExcelUIException("Must specify the same number of"
                                            + " MEF Values for each channel."
                                            + " Use 'None' to instruct FlowCal"
-                                           + " to ignore a detected peak.")
+                                           + " to ignore a detected"
+                                           + " subpopulation.")
             mef_values = np.array(mef_values)
 
             # Obtain standard curve transformation
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -470,6 +470,17 @@ def process_beads_table(beads_table,
                     mef = [int(e) if e.strip().isdigit() else np.nan
                            for e in mef]
                     mef_values.append(mef)
+
+            # Ensure matching number of `mef_values` for all channels (this
+            # implies that the calibration beads have the same number of
+            # subpopulations for all channels).
+            if mef_values:
+                if not np.all([len(mef_values_channel)==len(mef_values[0])
+                               for mef_values_channel in mef_values]):
+                    raise ExcelUIException("Must specify the same number of"
+                                           + " MEF Values for each channel."
+                                           + " Use 'None' to instruct FlowCal"
+                                           + " to ignore a detected peak.")
             mef_values = np.array(mef_values)
 
             # Obtain standard curve transformation
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -537,9 +537,11 @@ def get_transform_fxn(data_beads,
     ----------
     data_beads : FCSData object
         Flow cytometry data, taken from calibration beads.
-    mef_values : array
-        Known MEF values of the calibration beads' subpopulations, for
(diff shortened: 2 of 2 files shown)


JS3xton:
All unit tests pass.

To further test this, I modified `examples/experiment.xlsx`:

mef_vals_error

and ran the Excel UI:

mef_vals_error3

The resulting output file had the following Error message in the Analysis Notes column:

mef_vals_error2

FlowCal/excel_ui.py
```diff
@@ -470,6 +470,17 @@ def process_beads_table(beads_table,
...
+            # Ensure matching number of `mef_values` for all channels (this
+            # implies that the calibration beads have the same number of
+            # subpopulations for all channels).
+            if mef_values:
+                if not np.all([len(mef_values_channel)==len(mef_values[0])
+                               for mef_values_channel in mef_values]):
+                    raise ExcelUIException("Must specify the same number of"
+                                           + " MEF Values for each channel."
+                                           + " Use 'None' to instruct FlowCal"
+                                           + " to ignore a detected peak.")
```
castillohair: Replace "peak" by "subpopulation". I don't think we're referring to them as "peaks" anywhere else.

JS3xton: done in 5288de87b6631c05d4906d11ebe8d9f091f2f298

FlowCal/excel_ui.py
```diff
@@ -470,6 +470,18 @@ def process_beads_table(beads_table,
...
+            # implies that the calibration beads have the same number of
+            # subpopulations for all channels).
+            if mef_values:
+                if not np.all([len(mef_values_channel)==len(mef_values[0])
+                               for mef_values_channel in mef_values]):
+                    raise ExcelUIException("Must specify the same number of"
+                                           + " MEF Values for each channel."
+                                           + " Use 'None' to instruct FlowCal"
+                                           + " to ignore a detected"
+                                           + " subpopulation.")
```
castillohair, line 501: I realized that we're not being consistent with capitalization on exceptions in the Excel UI. We agreed that exception messages should start with lowercase; however some ExcelUIExceptions in this module start with lowercase, and some with uppercase. I guess uppercase is nice because this ends in a message in the Excel file.

FlowCal/mef.py
```diff
@@ -727,10 +729,16 @@ def get_transform_fxn(data_beads,
...
         mef_channels = [mef_channels]
-        mef_values = np.array([mef_values]).copy()
+        mef_values   = [mef_values]
+
+    # Ensure matching number of `mef_values` for all channels (this implies
+    # that the calibration beads have the same number of subpopulations for
+    # all channels).
+    if not np.all([len(mef_values_channel)==len(mef_values[0])
+                   for mef_values_channel in mef_values]):
+        raise ValueError("mef_values must be same length for all channels")
```
castillohair: I find the message a little confusing. It is not clear that it is the lengths of the innermost sequences in `mef_values` that should be the same.

castillohair:
Tested on example files (not only the Excel UI example but also the script), and confirmed that it functions as expected. I left some comments on the code. After addressing those, we should be ready to merge.

FlowCal/excel_ui.py
JS3xton: > We agreed that exception messages should start with lowercase; however some ExcelUIExceptions in this module start with lowercase, and some with uppercase.

true

> I guess uppercase is nice because this ends in a message in the Excel file.

agree

If you want to standardize this systematically across the entire module, I think that should be a separate issue (trivial as it is). I'm gonna assume you don't want any changes to this Pull Request, though.

FlowCal/mef.py
JS3xton: Attempted to clarify in 1505401303c7979ffa385c5a3f9a9fbdafd98ce3.

Issue #258, opened 2017-06-10
Anaconda installer does not add `python` to the PATH on windows


castillohair:
After running the Anaconda installer version 4.4.0 with the default options on windows, running `python` from the command line results in an "command not found" error. After reinstalling I noticed the following difference between the new installer and the old one:

Anaconda 4.3.1:
![image](https://user-images.githubusercontent.com/1529481/27006856-d3e41096-4e05-11e7-8b86-033f2498f142.png)

Anaconda 4.4.0:
![image](https://user-images.githubusercontent.com/1529481/27006849-a94d7958-4e05-11e7-903e-539b57b78709.png)

The top option is not enabled by default, and a scary red message appears when you click on it. The result of not enabling this option is that `python` is not added to the PATH. This results in our `.bat` scripts and examples not working.

The only thing we can do about this is to explicitly indicate in the installation instructions that both options should be enabled, despite the scary messages. Currently the instructions say:

![image](https://user-images.githubusercontent.com/1529481/27006869-249a988e-4e06-11e7-937e-3151243fccce.png)

I believe a good place to indicate this would be after step 2, in a very obnoxious highlighted rectangle. It would also be useful to verify whether this happens in OSX or not.

castillohair:
Confirmed the issue on another Windows machine. Also, it seems that this is not an issue on OSX, as the default is to add `python` to the path, and the option to not do it is not easy to access. I'll go ahead and modify the documentation as indicated above.

castillohair:
I changed the documentation to the following:

![screenshot from 2017-06-11 21-24-11](https://user-images.githubusercontent.com/1529481/27017238-b17a8606-4eec-11e7-9adc-44702fd130c7.png)

Pull request has been submitted.

Issue #230, opened 2017-01-08
Deprecation of GMM functions in scikit-learn


castillohair:
Apparently two of the functions that we use from `scikit-learn` will be deprecated soon. The following appears when running FlowCal 1.1.1 in linux with scikit-learn 0.18.1:

```
/home/sebastian/src/FlowCal-pypi/venv/local/lib/python2.7/site-packages/sklearn/utils/deprecation.py:52: DeprecationWarning: Class GMM is deprecated; The class GMM is deprecated in 0.18 and will be  removed in 0.20. Use class GaussianMixture instead.
  warnings.warn(msg, category=DeprecationWarning)
/home/sebastian/src/FlowCal-pypi/venv/local/lib/python2.7/site-packages/sklearn/utils/deprecation.py:70: DeprecationWarning: Function log_multivariate_normal_density is deprecated; The function log_multivariate_normal_density is deprecated in 0.18 and will be removed in 0.20.
  warnings.warn(msg, category=DeprecationWarning)
```

castillohair:
The whole GMM class is going to be dropped in favor of a new GaussianMixture class. Hopefully it is not too hard to replace. http://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html

castillohair:
Initial attempt was made in 8a6ff27d883e8dc9a39b64771181d31d7ac59fb3. Note that there is no one-to-one mapping between all the parameters in GMM and GaussianMixture, as far as I can tell. Also, it might be worth revising the initialization procedure since it may be completely ignored in the new class, which apparently forces the use of kmeans or random initialization. Also, it worries me that I haven't found a way to force the class to not update the weights, which is what we'be been doing so far.

Accepting a solution to this issue will require EXTENSIVE testing to make sure that the performance of the new clustering method is at least as good as the old one. The new method seems to work well on the example files, but this is only one data point.

Also, documentation will need to be updated in several places.

castillohair:
[...]

PR #259, opened 2017-06-10
Pandas compatibility fix

### LINKED ISSUE #257: Incompatibility with new version of pandas
With pandas 0.20.1 (the version included with the last Anaconda 4.4.0), the following error appears after running the example file through the Excel UI:

```
Calculating statistics for all samples...
Saving output Excel file...
  File "C:\Users\casti\Anaconda2\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Users\casti\Anaconda2\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 1490, in <module>
    hist_sheet=args.histogram_sheet)
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 1446, in run
    write_workbook(output_path, table_list)
  File "d:\castillohair\documents\rice\lab\programs\flowcal\FlowCal\excel_ui.py", line 177, in write_workbook
    format_module = pd.formats.format
AttributeError: 'module' object has no attribute 'formats'
```

The error happens in `excel_ui.write_workbook()`, specifically in the following block of code, aimed at changing the style of the headers in the excel files to be written:

```python
    # Modify default header format
    # Pandas' default header format is bold text with thin borders. Here we
    # use bold text only, without borders.
    # The format module is in pd.core.format in pandas<=0.18.0,
    # pd.formats.format in pandas>=0.18.1.
    try:
        format_module = pd.core.format
    except AttributeError as e:
        format_module = pd.formats.format
    old_header_style = format_module.header_style
    format_module.header_style = {"font": {"bold": True}}
```

The issue is that the attribute `formats`, used to modify header styles, has changed places in version 0.20. In fact, this has changed in the past already, as indicated by the comments in the block of code above.

- Location of `header_style` in version 0.19.2: https://github.com/pandas-dev/pandas/blob/v0.19.2/pandas/formats/format.py#L1623
- Location of `header_style` in version 0.20.1: https://github.com/pandas-dev/pandas/blob/v0.20.1/pandas/io/formats/excel.py#L33
- Location of `header_style` in version 0.20.2: https://github.com/pandas-dev/pandas/blob/v0.20.2/pandas/io/formats/excel.py#L33

The solution to this issue should be straighforward given that we know the new location. However, since the `pandas` team seem to change this from time to time, and this block of code is only there for aesthetic reasons, I think it should be possible to wrap the whole thing in a `try` block, and simply using the default header style from pandas if it fails.

The resulting block should look something like this:

```python
try:
    if packaging.version.parse(pandas.__version__) \
            < packaging.version.parse('0.18'):
        format_module = pandas.core.format
    elif packaging.version.parse(pandas.__version__) \
            < packaging.version.parse('0.20'):
        format_module = pandas.formats.format
    else:
        import pandas.io.formats.excel
        format_module = pandas.io.formats.excel

    format_module.header_style = {"font": {"bold": True,
                                           "color": "FF1F497D"}}
except:
    pass
```

I'm using similar code in another package of mine. Also, the `packaging` module provides a nicer way to check for module versions, and it will be the way to go when solving #230.

castillohair:
Fixes #257. Adds the requirement for the `packaging` module, which is, as far as I've found, the best way to compare module versions. This will be useful later when fixing #230.

I intend the solution of this and #258 to be part of a hotfix, to be pushed ASAP.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -72,6 +72,8 @@ ignored by ``FlowCal``.
 import collections
 import os
 import os.path
+import packaging
+import packaging.version
 import platform
 import re
 import subprocess
@@ -170,13 +172,27 @@ def write_workbook(filename, table_list, column_width=None):
     # Pandas' default header format is bold text with thin borders. Here we
     # use bold text only, without borders.
     # The format module is in pd.core.format in pandas<=0.18.0,
-    # pd.formats.format in pandas>=0.18.1.
+    # pd.formats.format in pandas>=0.18.1, and pd.io.formats.excel in
+    # pandas>=0.20.
+    # Also, wrap in a try-except block in case format module is not found.
+    format_module_found = False
     try:
-        format_module = pd.core.format
+        # Get format module
+        if packaging.version.parse(pd.__version__) \
+                <= packaging.version.parse('0.18'):
+            format_module = pd.core.format
+        elif packaging.version.parse(pd.__version__) \
+                < packaging.version.parse('0.20'):
+            format_module = pd.formats.format
+        else:
+            import pandas.io.formats.excel as format_module
+        # Save previous style, replace, and indicate that previous style should
+        # be restored at the end
+        old_header_style = format_module.header_style
+        format_module.header_style = {"font": {"bold": True}}
+        format_module_found = True
     except AttributeError as e:
-        format_module = pd.formats.format
-    old_header_style = format_module.header_style
-    format_module.header_style = {"font": {"bold": True}}
+        pass
 
     # Generate output writer object
     writer = pd.ExcelWriter(filename, engine='xlsxwriter')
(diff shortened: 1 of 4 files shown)


FlowCal/excel_ui.py
```diff
@@ -170,13 +172,27 @@ def write_workbook(filename, table_list, column_width=None):
     # Pandas' default header format is bold text with thin borders. Here we
     # use bold text only, without borders.
     # The format module is in pd.core.format in pandas<=0.18.0,
-    # pd.formats.format in pandas>=0.18.1.
+    # pd.formats.format in pandas>=0.18.1, and pd.io.formats.excel in
+    # pandas>=0.20.
```
JS3xton: I think your intervals are overlapping in this description (e.g. v0.20.2 is both >= 0.18.1 and >= 0.20), but that's really not very important. The code appears to handle it correctly? (having not actually tested it myself)

JS3xton (approved):

FlowCal/excel_ui.py
castillohair: Lol yeah the code works. It has been tested with 0.18.0, 0.18.1,, 0.19.1, 0.20.1, and 0.20.2. I will update the description.

Issue #261, opened 2017-06-13
Documentation of `mef_outputs` in `excel_ui.process_beads_table()` is insufficient.


castillohair:
See #232 for more details, although the title is self-contained.

castillohair:
Merged in #262.

Issue #232, opened 2017-02-01
mef.get_transform_fxn() outputs should be OrderedDict instead of lists to carry around channel ID


JS3xton:
`mef.get_transform_fxn()` loops through specified MEF channels and builds a `functools.partial` function which transforms RFI values to MEF values based on specified bead data. If you ask for `full_output`, this function accumulates useful intermediate data structures and information for each specified MEF channel (specified via `mef_channels` input) in a series of lists and returns these lists to the user. E.g. from the `mef.get_transform_fxn()` documentation -
```
    fitting : dict, only if ``full_output==True``
        Results of the model fitting step, containing the following fields:
        std_crv : list
            Functions encoding the standard curves, for each channel in
            `mef_channels`.
        beads_model : list
            Functions encoding the fluorescence model of the
            calibration beads, for each channel in `mef_channels`.
        beads_params : list
            Fitted parameters of the bead fluorescence model, for each
            channel in `mef_chanels`.
        beads_model_str : list
            String representation of the bead models used, for each channel
            in `mef_channels`.
        beads_params_names : list
            Names of the parameters given in `beads_params`, for each
            channel in `mef_channels`.
```

I think it would be more useful if these accumulation data structures were `OrderedDict`s keyed on the `mef_channels` values. This seems like a natural way to carry around the channel ID for use in subsequent computations, and I believe the original list can be recovered easily via `OrderedDict.values()`.

My use case is that I'm using the `excel_ui.process_beads_table()` function from the Python API, and I want to print the detector voltages of the channels I've chosen to MEF.

`excel_ui.process_beads_table()` returns:
```
    beads_samples : list of FCSData objects
        A list of processed, gated, and transformed samples, as specified
        in `beads_table`, in the order of ``beads_table.index``.
    mef_transform_fxns : OrderedDict
        A dictionary of MEF transformation functions, indexed by
        ``beads_table.index``.
    mef_outputs : list
        A list with intermediate results of the generation of the MEF
        transformation functions, indexed by ``beads_table.index``. Only
        included if `full_output` is True.
```

[...]

PR #262, opened 2017-06-16
Expanded documentation of mef_outputs in excel_ui.process_beads_table().

castillohair:
Solves #261.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -307,14 +307,14 @@ def process_beads_table(beads_table,
     mef_transform_fxns : OrderedDict
         A dictionary of MEF transformation functions, indexed by
         ``beads_table.index``.
-    mef_outputs : list
+    mef_outputs : list, only if ``full_output==True``
         A list with intermediate results of the generation of the MEF
-        transformation functions. More precisely, each element of this list
-        is the unmodified output of calling ``mef.get_transform_fxn()``
-        with ``full_output=True``, for each processed and gated beads
-        sample. Each element of `mef_outputs` is a named tuple, as
-        specified in the documentation of ``mef.get_transform_fxn()``.
-        `mef_outputs` is only included if `full_output` is True.
+        transformation functions. More precisely, for every processed and
+        gated beads sample, :func:`FlowCal.mef.get_transform_fxn()` is
+        called with ``full_output=True``, and the full output (a
+        ``namedtuple``) is appended to this list. Please refer to the
+        output section of :func:`FlowCal.mef.get_transform_fxn()`'s
+        documentation for more information.
 
     """
     # Initialize output variables
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -297,8 +297,12 @@ def process_beads_table(beads_table,
         ``beads_table.index``.
     mef_outputs : list
         A list with intermediate results of the generation of the MEF
-        transformation functions, indexed by ``beads_table.index``. Only
-        included if `full_output` is True.
+        transformation functions. More precisely, each element of this list
+        is the unmodified output of calling ``mef.get_transform_fxn()``
+        with ``full_output=True``, for each processed and gated beads
+        sample. Each element of `mef_outputs` is a named tuple, as
+        specified in the documentation of ``mef.get_transform_fxn()``.
+        `mef_outputs` is only included if `full_output` is True.
 
     """
     # Initialize output variables
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -309,10 +309,11 @@ def process_beads_table(beads_table,
         ``beads_table.index``.
     mef_outputs : list, only if ``full_output==True``
         A list with intermediate results of the generation of the MEF
-        transformation functions. More precisely, for every processed and
-        gated beads sample, :func:`FlowCal.mef.get_transform_fxn()` is
-        called with ``full_output=True``, and the full output (a
-        ``namedtuple``) is appended to this list. Please refer to the
+        transformation functions. For every entry in `beads_table`,
+        :func:`FlowCal.mef.get_transform_fxn()` is called on the
+        corresponding processed and gated beads sample with
+        ``full_output=True``, and the full output (a `MEFOutput`
+        ``namedtuple``) is appended to `mef_outputs`. Please refer to the
         output section of :func:`FlowCal.mef.get_transform_fxn()`'s
         documentation for more information.
 
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -579,9 +579,9 @@ def get_transform_fxn(data_beads,
         Flow cytometry data, taken from calibration beads.
     mef_values : sequence of sequences
         Known MEF values for the calibration bead subpopulations, for each
-        channel specified in `mef_channels`. The innermost sequences must have
-        the same length (the same number of bead subpopulations must exist for
-        each channel).
+        channel specified in `mef_channels`. The innermost sequences must
+        have the same length (the same number of bead subpopulations must
+        exist for each channel).
     mef_channels : int, or str, or list of int, or list of str
         Channels for which to generate transformation functions.
     verbose : bool, optional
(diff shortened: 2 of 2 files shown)


JS3xton:
I struggled with the documentation for `mef.get_transform_fxn()` in addition to `excel_ui.process_beads_table()`.

### `excel_ui.process_beads_table()` Documentation

(note: `beads_samples` and `mef_outputs` may eventually be `OrderedDict`s, pending #232)

Portion of proposed documentation:
```
Returns
-------
beads_samples : list of FCSData objects
    A list of processed, gated, and transformed samples, as specified
    in `beads_table`, in the order of ``beads_table.index``.
mef_transform_fxns : OrderedDict
    A dictionary of MEF transformation functions, indexed by
    ``beads_table.index``.
mef_outputs : list
    A list with intermediate results of the generation of the MEF
    transformation functions. More precisely, each element of this list
    is the unmodified output of calling ``mef.get_transform_fxn()``
    with ``full_output=True``, for each processed and gated beads
    sample. Each element of `mef_outputs` is a named tuple, as
    specified in the documentation of ``mef.get_transform_fxn()``.
    `mef_outputs` is only included if `full_output` is True.
```

* `namedtuple` should not have a space in it
* I don't think you should say "`mef_outputs` is only included if `full_output` is True.". You should be consistent with our other documentation of `full_output` parameters, which appears to be more like this (i.e. mention "only if ``full_output==True``" at the top, not in the parameter description):

```
mef_channels : int, or str, or list, only if ``full_output==True``
    Channels on which transformation functions have been generated.
    Directly copied from the `mef_channels` argument.
```
* You can say

```
"Each element of `mef_outputs` is a named tuple, as
    specified in the documentation of ``mef.get_transform_fxn()``."
```

but I think you should also reproduce the description of this data structure from `mef.get_transform_fxn()` so the user doesn't have to chase that down. Going to another function description is certainly better than having to read the source code, but I still think it's annoying for someone just trying to use the `excel_ui.process_beads_table()` function and who isn't interested in the underlying implementation details (e.g. someone who was previously using the Excel UI and their use of `excel_ui.process_beads_table()` represents their first foray into the Python API). It would require duplicating some documentation, though, (which would need to be updated in two places if it ever changes), which may be undesirable, but at present the outputs of this function are uninterpretable without combing through the code, in my opinion. This also gives you the opportunity to more explicitly describe the `namedtuple`, which is not explicitly described in the documentation for `mef.get_transform_fxn()`. So maybe something like:

```
mef_outputs : list, only if ``full_output==True``
    A list with intermediate results of the generation of the MEF
    transformation functions.

    Each list element is a MEFOutput namedtuple that is the unmodified output
    of calling ``mef.get_transform_fxn()`` with ``full_output=True`` on the processed
    and gated beads sample associated with each row in `beads_table`. The fields
    of the MEFOutput namedtuple are:

    transform_fxn : function
        Transformation function to convert flow cytometry data from RFI
        units to MEF. This function has the same basic signature as the
        general transformation function specified in ``FlowCal.transform``.
    mef_channels : int, or str, or list
        Channels on which transformation functions have been generated.
    clustering : dict
        Results of the clustering step, containing the following fields:
        labels : array
            Labels for each event in `data_beads`.
    statistic : dict
        Results of the calculation of bead subpopulations' fluorescence,
        containing the following fields:
        values : list
            The representative fluorescence values of each
            subpopulation, for each channel in `mef_channels`.
    selection : dict
        Results of the subpopulation selection step, containing the
        following fields:
        rfi : list
            The fluorescence values of each selected subpopulation in
            RFI units, for each channel in `mef_channels`.
        mef : list
            The fluorescence values of each selected subpopulation in
            MEF units, for each channel in `mef_channels`.
    fitting : dict
        Results of the model fitting step, containing the following fields:
        std_crv : list
            Functions encoding the standard curves, for each channel in
            `mef_channels`.
        beads_model : list
            Functions encoding the fluorescence model of the
            calibration beads, for each channel in `mef_channels`.
        beads_params : list
            Fitted parameters of the bead fluorescence model, for each
            channel in `mef_chanels`.
        beads_model_str : list
            String representation of the bead models used, for each channel
            in `mef_channels`.
        beads_params_names : list
            Names of the parameters given in `beads_params`, for each
            channel in `mef_channels`.

     (see mef.get_transform_fxn() documentation for more details on the MEFOutput namedtuple)
```

* I think it would be very helpful to have an example in the documentation that illustrates what `mef_outputs` looks like if `excel_ui.process_beads_table()` is called on the example file. So for

```python
import FlowCal as fc
i_table = fc.excel_ui.read_table('experiment.xlsx', 'Instruments', index_col='ID')
b_table = fc.excel_ui.read_table('experiment.xlsx', 'Beads', index_col='ID')
beads_samples, mef_transform_fxns, mef_outputs = fc.excel_ui.process_beads_table(b_table, i_table, full_output=True)
```

the `mef_outputs` structure looks like:
```
[
    FlowCal.mef.MEFOutput(
        transform_fxn=functools.partial,
        mef_channels=[u'FL1'],
        clustering={
            'labels' : [7,2,1, ... 4, 3, 5]
        },
        statistic={
            'values' : [np.array([100, 149, 230, 433, 1240, 3105, 7773, 9305])]
        },
        selection={
            'rfi' : [np.array([100, 149, 230, 433, 1240, 3105, 7773])]
            'mef' : [np.array([0,646,1704,4827,15991, 47609, 135896])]
        },
        fitting={
            'std_crv' : [lambda function]
            'beads_model' : [lambda function]
            'beads_params' : [np.array(1.09e0, 2.02e0, 1.15e3)]
            'beads_model_str' : ['m*log(fl_rfi) + b = log(fl_mef_auto + fl_mef)']
            'beads_params_names' : [['m', 'b', 'fl_mef_auto]]
        },
    )
]
```

(this illustration might be more useful if there was more than 1 entry in the Beads table of the example file)

### `mef.get_transform_fxn()` Documentation

Partial current documentation:

```
    Returns
    -------
    transform_fxn : function
        Transformation function to convert flow cytometry data from RFI
        units to MEF. This function has the same basic signature as the
        general transformation function specified in ``FlowCal.transform``.
    mef_channels : int, or str, or list, only if ``full_output==True``
        Channels on which transformation functions have been generated.
        Directly copied from the `mef_channels` argument.
    clustering : dict, only if ``full_output==True``
        Results of the clustering step, containing the following fields:
        labels : array
            Labels for each event in `data_beads`.
    statistic : dict, only if ``full_output==True``
        Results of the calculation of bead subpopulations' fluorescence,
        containing the following fields:
        values : list
            The representative fluorescence values of each
            subpopulation, for each channel in `mef_channels`.
    selection : dict, only if ``full_output==True``
        Results of the subpopulation selection step, containing the
        following fields:
        rfi : list
            The fluorescence values of each selected subpopulation in
            RFI units, for each channel in `mef_channels`.
        mef : list
            The fluorescence values of each selected subpopulation in
            MEF units, for each channel in `mef_channels`.
    fitting : dict, only if ``full_output==True``
        Results of the model fitting step, containing the following fields:
        std_crv : list
            Functions encoding the standard curves, for each channel in
            `mef_channels`.
        beads_model : list
            Functions encoding the fluorescence model of the
            calibration beads, for each channel in `mef_channels`.
        beads_params : list
            Fitted parameters of the bead fluorescence model, for each
            channel in `mef_chanels`.
        beads_model_str : list
            String representation of the bead models used, for each channel
            in `mef_channels`.
        beads_params_names : list
            Names of the parameters given in `beads_params`, for each
            channel in `mef_channels`.
```
* for `transform_fxn`, saying "This function has the same basic signature as the
        general transformation function specified in ``FlowCal.transform``." is fine, but again, I would reproduce that function signature here so the user doesn't have to chase that down.
* Shouldn't `clustering['labels']` also indicate that it's "for each channel in `mef_channels`"?
* How the channels dimensions is represented in these data structures is still not very clearly presented in my opinion (i.e. the lists in the `clustering`, `statistic`, `selection`, and `fitting` `dict`s). Similar to above, an example illustrated in the documentation could help to illustrate this.

(any changes to `mef.get_transform_fxn()` should be propogated to any duplicated documentation mentioned above in `excel_ui.process_beads_table()`)

JS3xton (changes requested):
See posted comment.

castillohair:
I'll start by saying that I'm strongly against duplicating pieces of documentation. This will make it really hard to maintain if we want to make a change. Updating documentation and examples is already difficult because we have these in docstrings, `.rst` files with layman explanations and tutorials, and example python scripts. Duplicating information across docstrings will only make it harder, and it is unnecessary if we can effectively redirect the user to the relevant documentation portion.

Fortunately, there seems to be a good middle ground. There is a way to make [links to other parts of the documentation](http://www.sphinx-doc.org/en/stable/domains.html#cross-referencing-python-objects). This will allow the user to be one click away from `mef.get_transform_fxn()` when reading `process_beads_table()`'s documentation. The following screenshot shows how it looks:

screenshot from 2017-12-29 23-08-33

(the pointer looks like a hand with its finger out, like when it's on top of a hyperlink, but I don't know why the screenshot program doesn't capture it). It is a hyperlink, which takes the user to `mef.get_transform_fxn()`'s documentation as intended. I don't like the color, but it is a problem of the website theme that we're currently using, and I'd rather not spend my time modifying themes.

If we agree that this is sufficient, most of your recent complains should then be fixed in `mef.get_transform_fxn()`'s docstring. As of b503b95dc52689d566f979939b2ea7f080d48d04, this docstring has been modified as follows:
- Documentation of each output element has been expanded to make it more clear that most elements are dictionaries, that each element below is a key:value pair, that these values are usually lists and what type of elements these lists include, and the dimensions of these lists and the elements inside.
- Some examples have been added at the end, one of which shows an example full output. This is somewhat based on your example.

I think the expanded description of the output of `mef.get_transform_fxn()`, combined with the addition of a clickable link in `process_beads_table()`'s documentation to `mef.get_transform_fxn()`, should be enough to address all of your concerns.

FlowCal/excel_ui.py
```diff
@@ -307,10 +307,14 @@ def process_beads_table(beads_table,
...
+    mef_outputs : list, only if ``full_output==True``
         A list with intermediate results of the generation of the MEF
-        transformation functions, indexed by ``beads_table.index``. Only
-        included if `full_output` is True.
+        transformation functions. More precisely, for every processed and
+        gated beads sample, :func:`FlowCal.mef.get_transform_fxn()` is
+        called with ``full_output=True``, and the full output (a
+        ``namedtuple``) is appended to this list. Please refer to the
+        output section of :func:`FlowCal.mef.get_transform_fxn()`'s
+        documentation for more information.
```
JS3xton: > "for every processed and gated beads sample"

is a little kludge to understand. Maybe more simply/directly:

> "For every entry in ``beads_table``, :func:`FlowCal.mef.get_transform_fxn()` is called on the entry's processed and gated beads sample with ..."

This is minor, though.

JS3xton: Might also say `mef_outputs : list of MEFOutput, only if ...` (add "of MEFOutput")

FlowCal/mef.py
```diff
@@ -597,64 +597,112 @@ def get_transform_fxn(data_beads,
         Name to use for plot files. If None, use ``str(data_beads)``.
     full_output : bool, optional
         Flag specifying whether to include intermediate results in the
-        output. If `full_output` is True, the function returns a named
-        tuple with fields as described below. If `full_output` is False,
-        the function only returns the calculated transformation function.
+        output. If `full_output` is True, the function returns a
+        ``namedtuple`` with fields as described below. If `full_output` is
+        False, the function only returns the calculated transformation
+        function.
```
JS3xton: Include the name of the `namedtuple` so we brand it consistently. Will also give users a handle to look for if they're coming over from the documentation for `excel_ui.process_beads_table()`:

"If `full_output` is True, the function returns a **MEFOutput** ``namedtuple`` with fields as described below."

FlowCal/excel_ui.py
JS3xton: > "and the full output (a **MEFOutput** ``namedtuple``)"

(add "MEFOutput" again)

FlowCal/mef.py
```diff
@@ -597,64 +597,112 @@ def get_transform_fxn(data_beads,
...
 
-        labels : array
-            Labels for each event in `data_beads`.
+        "labels" : array
+            Array of length ``N``, where ``N`` is the number of events in
+            `data_beads`. This array contains labels indicating which
+            subpopulation each event has been assigned to by the clustering
+            algorithm. Labels range from ``0`` to ``M - 1``, where ``M`` is
+            the number of MEF values specified, and therefore the number of
+            subpopulations identified by the clustering algorithm.
```
JS3xton, line 631: I don't understand what the `keys` are for the `clustering` `dict`.

FlowCal/mef.py
```diff
@@ -597,64 +597,112 @@ def get_transform_fxn(data_beads,
...
 
-        values : list
-            The representative fluorescence values of each
-            subpopulation, for each channel in `mef_channels`.
+        "values" : list of arrays
+            Each array contains the representative fluorescence values of
+            all subpopulations, for a specific fluorescence channel from
+            `mef_channels`. Therefore, each array has a length equal to the
+            number of subpopulations, and the outer list has as many arrays
+            as the number of channels in `mef_channels`.
```
JS3xton, line 646: I don't understand what the `keys` are for the `statistic` `dict`.

FlowCal/mef.py
JS3xton: Especially confusing because you even mention "key : value" pairs, but then proceed to only list a "value" (the "labels" array?). Needless to say, I'm confused.

JS3xton: Oh wait, "labels" is the `key`? and the `array` is the value? And `clustering` just happens to only have 1 `key`? OK that makes more sense, but it's not communicated clearly here.

JS3xton: Maybe something as simple as:

`selection = {'labels' : np.array}`

before your descriptions of the key:value pairs. I'm still confused by your initial description. Maybe "The structure of this dictionary is:", and then have:

`selection = {'labels' : np.array}`

And then go with the descriptions you already have, because those are useful once you understand what they're describing.

FlowCal/mef.py
JS3xton: Similar to the `selection` dictionary, maybe remove "This dictionary contains the following ``key : value`` pairs:" and replace it with

"The structure of this dictionary is:

`statistic = {'values' : [np.array, ...]}`"

FlowCal/mef.py
```diff
@@ -597,64 +597,112 @@ def get_transform_fxn(data_beads,
...
+            Each array contains the fluorescence values of each selected
+            subpopulation in MEF units, for a specific fluorescence channel
+            from `mef_channels`. The outer list has as many arrays as the
+            number of channels in `mef_channels`. Because the selection
+            step may discard subpopulations, each array has a length less
+            than or equal to the total number of subpopulations.
+            Furthermore, different arrays in this list may not have the
+            same length. However, the length of each array is consistent
+            with the corresponding array in ``selection["rfi"]`` (see
+            above).
```
JS3xton, line 678: Consider replacing "This dictionary contains the following ``key : value`` pairs:"

with

"The structure of this dictionary is:

`selection = {'rfi' : [np.array,...], 'mef' : [np.array,...]}`" (<-- probably make the dictionary definition span 2 lines)

FlowCal/mef.py
```diff
@@ -597,64 +597,112 @@ def get_transform_fxn(data_beads,
...
+            after applying it to the MEF and RFI fluorescence values of a
+            specific channel from `mef_channels` .
+        "beads_params_names" : list of list
             Names of the parameters given in `beads_params`, for each
-            channel in `mef_channels`.
+            channel in `mef_channels`. Each element of this list is the
+            ``beads_params_names`` output of the fitting function (see
+            required signature of the ``fitting_fxn`` optional parameter),
+            after applying it to the MEF and RFI fluorescence values of a
+            specific channel from `mef_channels` .
```
JS3xton, line 726: Consider replacing "This dictionary contains the following ``key : value`` pairs:"

with

"The structure of this dictionary is:

```python
fitting = {
    'std_crv'            : [func,...],
    'beads_model'        : [func,...],,
    'beads_params'       : [np.array,...],
    'beads_model_str'    : ['str',...],
    'beads_params_names' : [[],...],
}
```

FlowCal/mef.py
```diff
@@ -752,6 +801,79 @@ def get_transform_fxn(data_beads,
...
+
+    >>> transform_fxn = FlowCal.mef.get_transform_fxn(
+    ...     beads_data,
+    ...     mef_channels=['FL1', 'FL3'],
+    ...     mef_values=[np.array([    0,   646,   1704,   4827,
+    ...                           15991, 47609, 135896, 273006],
+    ...                 np.array([    0,  1614,   4035,   12025,
+    ...                           31896, 95682, 353225, 1077421]],
+    ...     )
+    >>> sample_mef = transform_fxn(sample_rfi, ['FL1', 'FL3'])
```
JS3xton: I prefer including the parameter names to be as specific as possible, e.g.

`>>> sample_mef = transform_fxn(data_rfi=sample_rfi, channels=['FL1', 'FL3'])`

Up to you, though.

FlowCal/mef.py
```diff
@@ -752,6 +801,79 @@ def get_transform_fxn(data_beads,
...
+    ...     beads_data,
+    ...     mef_channels=['FL1', 'FL3'],
+    ...     mef_values=[np.array([    0,   646,   1704,   4827,
+    ...                           15991, 47609, 135896, 273006],
+    ...                 np.array([    0,  1614,   4035,   12025,
+    ...                           31896, 95682, 353225, 1077421]],
+    ...     full_output=True)
+
+    In this case, the output ``get_transform_output`` will be a
+    ``namedtuple`` similar to the following::
```
JS3xton: explicitly state "**MEFOutput** ``namedtuple``".

JS3xton (approved):
- Linking documentation = 👍
- The example in `mef.get_transform_fxn()` is helpful.
- In `mef.get_transform_fxn()` documentation for `data_beads` input, maybe: "Flow cytometry data DESCRIBING calibration beads." (instead of "taken from"). This is minor, though.

Issue #249, opened 2017-04-25
Error plotting density2d


JS3xton:
I get an error on a particular FCS file when I try to generate a 2D density plot.

```python
import FlowCal as fc
import matplotlib.pyplot as plt
d = fc.io.FCSData('density_plot_fail.fcs')
fc.plot.density2d(d, channels=['FSC-A', 'SSC-A'])
plt.show()
```

yields

```
Traceback (most recent call last):
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\backends\backend_qt5agg.py", line 197, in __draw_idle_agg
    FigureCanvasAgg.draw(self)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\backends\backend_agg.py", line 464, in draw
    self.figure.draw(self.renderer)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\artist.py", line 63, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\figure.py", line 1143, in draw
    renderer, self, dsu, self.suppressComposite)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\image.py", line 139, in _draw_list_compositing_images
    a.draw(renderer)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\artist.py", line 63, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\axes\_base.py", line 2409, in draw
    mimage._draw_list_compositing_images(renderer, self, dsu)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\image.py", line 139, in _draw_list_compositing_images
    a.draw(renderer)
  File "C:\Users\sexto\Anaconda2\envs\FlowCalv1.1.2\lib\site-packages\matplotlib\artist.py", line 63, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
[...]

PR #263, opened 2017-06-18
Plot logicle fix

castillohair:
Solves #249.

### DIFF
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -435,22 +435,27 @@ class _LogicleLocator(matplotlib.ticker.Locator):
         """
         # Extract base from transform object
         b = self._transform.base
-        # The domain is divided into two sections, only some of which may
-        # actually be present.
+        # The logicle domain is divided into two regions: A "linear" region,
+        # which may include negative numbers, and a "logarithmic" region, which
+        # only includes positive numbers. These two regions are separated by a
+        # value t, given by the logicle equations. An illustration is given
+        # below.
         #
         # -t ==0== t ========>
-        #    bbbbb   ccccccccc
+        #     lin       log
         #
-        # Where ``t`` is obtained from the logicle linear range.
+        # vmin and vmax can be anywhere in this domain, meaning that both should
+        # be greater than -t.
         #
-        # c) will have ticks at integral log positions. The number of ticks
-        # needs to be reduced if there are more than self.numticks of them.
-        #
-        # b) will have one tick at zero. Subticks will be added from the
-        # smallest integral log in c to the next smallest, all throughout
-        # b. The negative part of b will have similar subticks.
+        # The logarithmic region will only have major ticks at integral log
+        # positions. The linear region will have a major tick at zero, and one
+        # major tick at the largest absolute  integral log value in screen
+        # inside this region. Subticks will be added at multiples of the
+        # integral log positions.
 
         # If the linear range is too small, create new transformation object
+        # with slightly wider linear range. Otherwise, the number of decades
+        # below will be infinite
         if self._transform.W == 0 or \
                 self._transform.M / self._transform.W > self.numticks:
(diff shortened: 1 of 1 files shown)


castillohair:
This pull request has code that improves function of the Logicle locator (the class that decides where to place the tickers). Previously, the locator could not handle cases in which only the "linear" region of the logicle scale was in screen. This led to failure when zooming into the linear region of a plot, and complete failure when opening a file in which negative events have a greater absolute value than positive events (because, in this case, all of the plot is in the "linear" region).

As for a7b39bf7db54f8fd5b296310d5caa79d335eab5a, the two files in #249 can now be plotted. The following are the density maps of forward/side scatter for both files:

File 1:
fail001

File 2:
fail002

To test this even further, I wrote the following code:

```python
import FlowCal as fc
import matplotlib.pyplot as plt

files = ['data_moake.fcs',
         'data_sea.fcs',
         'density_plot_fail.fcs']
channels = ['SSC', 'FSC-A', 'SSC-A']
xlims = [[None, [10, 200], [0, 200], [1e-3, 200],
          [0, 100], [0, 1], [0.02, 0.09], [50, 110]],
         [None, [-1e3, 2e5], [0, 2e5], [1e-3, 2e5],
          [-1e3, 1e3], [-4e3, -1e3], [-4e3, -990], [-4e3, 1e2]],
         [None, [-1e3, 1e4], [0, 1e4], [1e3, 1e4],
          [-2e4, -1e2], [-2e4, -5e3], [-2e4, 1e3], [1e5, 1e6]]]

for i, (file, channel, xlim) in enumerate(zip(files, channels, xlims)):
    print "File {}: {}...".format(i+1, file)
    d = fc.io.FCSData(file)
    d = fc.transform.to_rfi(d)

    plt.figure(figsize=(8, 10))
    for xi, xl in enumerate(xlim):
        print "   Zoom level {}...".format(xi + 1)
        plt.subplot(4, 2, xi + 1)
        fc.plot.hist1d(d, channel=channel)
        if xl is None:
            plt.title('Default')
        else:
            plt.title(str(xl))
        plt.xlim(xl)
    plt.tight_layout()
    plt.savefig('File{:03d}_zoom.png'.format(i+1), dpi=200)
    plt.close()
```

And tested it against [these files](https://github.com/taborlab/FlowCal/files/1082863/logicle_test.zip). I think this example code makes a good job a highlighting the improved abilities of the logicle locator, as well as its current limitations. Obviously, the code doesn't run without this fix.

For file 1 (Moake lab flow cytometer):
file001_zoom

Because of the decision to only place text under ticks that have an integral decade value (i.e. 10^a, where a is an integer), it is not hard to get into a situation like in subfigure 7, in which the zoom is such that no tick value is shown. This can also happen when using default matplotlib scaling, although it is a little harder to do. A quick solution from the user's perspective is to zoom out a little bit. Trying to place text on ticks with non-integral decade values can result in serious text agglomeration issues. I think fully resolving this would require me to implement a custom [Formatter](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/ticker.py), which is outside the scope of this pull request, and would require way more time that I have right now.

For file 2 (Moake lab's flow cytometer):
file002_zoom

Same issue in subplot 6. Again, the user can avoid this by zooming out a little (see subplot 7).

For file 3 (first file that failed to load in #249):
file003_zoom

In summary, this pull request completely solves issue #249. It seems that placement of text under ticks is not perfect, but it is not inferior to some native matplotlib scales, and therefore I will not be improving it further in this pull request. Nevertheless, this is a big improvement over the current `develop` branch.

JS3xton (approved):
I can't vouch for the technical details because I haven't taken the time to study the implementation closely, but the tests you've shown convince me that it's definitely better than it was. I also didn't see any glaring errors/typos from skimming the code.

Issue #264, opened 2017-06-18
Python 3 support


castillohair:
`FlowCal` should be fully compatible with Python 3, obviously without dropping support for Python 2.7.

JS3xton:
Previously discussed in #100.

PR #265, opened 2017-06-18
Python3 Support

castillohair:
Solves #264. Tested on python 3.5 and 3.6. Unit tests pass, with a few warning messages (same as current develop (893acba3378e493829712a625e80758024bd60aa) with python 2.7). Example files run through the Excel UI and produce the appropriate output.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -77,10 +77,18 @@ import packaging.version
 import platform
 import re
 import subprocess
+import sys
 import time
-from Tkinter import Tk
-from tkFileDialog import askopenfilename
 import warnings
+
+# Tkinter is imported differently depending on which version of python we're
+# using.
+if sys.version_info.major == 2:
+    from Tkinter import Tk
+    from tkFileDialog import askopenfilename
+elif sys.version_info.major == 3:
+    from tkinter import Tk
+    from tkinter.filedialog import askopenfilename
 
 from matplotlib import pyplot as plt
 import numpy as np
@@ -133,7 +141,8 @@ def read_table(filename, sheetname, index_col=None):
 
     """
     # Catch sheetname as list or None
-    if sheetname is None or hasattr(sheetname, '__iter__'):
+    if sheetname is None or \
+        (hasattr(sheetname, '__iter__') and not isinstance(sheetname, str)):
         raise TypeError("sheetname should specify a single sheet")
 
     # Load excel table using pandas
@@ -205,7 +214,7 @@ def write_workbook(filename, table_list, column_width=None):
         df.to_excel(writer, sheet_name=sheet_name, index=False)
         # Set column width
         if column_width is None:
-            for i, (col_name, column) in enumerate(df.iteritems()):
+            for i, (col_name, column) in enumerate(iter(df.items())):
                 # Get the maximum number of characters in a column
                 max_chars_col = column.astype(str).str.len().max()
                 max_chars_col = max(len(col_name), max_chars_col)
@@ -1287,7 +1296,7 @@ def generate_about_table(extra_info={}):
     keywords.append('Time of analysis')
     values.append(time.strftime("%I:%M:%S%p"))
     # Add additional keyword:value pairs
-    for k, v in extra_info.items():
+    for k, v in iter(extra_info.items()):
         keywords.append(k)
         values.append(v)
 
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -76,6 +76,7 @@ import packaging
 import packaging.version
 import platform
 import re
+import six
 import subprocess
 import sys
 import time
@@ -142,7 +143,8 @@ def read_table(filename, sheetname, index_col=None):
     """
     # Catch sheetname as list or None
     if sheetname is None or \
-        (hasattr(sheetname, '__iter__') and not isinstance(sheetname, str)):
+            (hasattr(sheetname, '__iter__') \
+            and not isinstance(sheetname, six.string_types)):
         raise TypeError("sheetname should specify a single sheet")
 
     # Load excel table using pandas
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -78,16 +78,17 @@ import platform
 import re
 import six
 import subprocess
-import sys
 import time
 import warnings
 
 # Tkinter is imported differently depending on which version of python we're
 # using.
-if sys.version_info.major == 2:
+# six.PY2 is True when the code is running in python 2, False otherwise.
+# six.PY3 is the equivalent for python 3.
+if six.PY2:
     from Tkinter import Tk
     from tkFileDialog import askopenfilename
-elif sys.version_info.major == 3:
+elif six.PY3:
     from tkinter import Tk
     from tkinter.filedialog import askopenfilename
 
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -217,7 +217,7 @@ def write_workbook(filename, table_list, column_width=None):
         df.to_excel(writer, sheet_name=sheet_name, index=False)
         # Set column width
         if column_width is None:
-            for i, (col_name, column) in enumerate(iter(df.items())):
+            for i, (col_name, column) in enumerate(six.iteritems(df)):
                 # Get the maximum number of characters in a column
                 max_chars_col = column.astype(str).str.len().max()
                 max_chars_col = max(len(col_name), max_chars_col)
@@ -1299,7 +1299,7 @@ def generate_about_table(extra_info={}):
     keywords.append('Time of analysis')
     values.append(time.strftime("%I:%M:%S%p"))
     # Add additional keyword:value pairs
-    for k, v in iter(extra_info.items()):
+    for k, v in six.iteritems(extra_info):
         keywords.append(k)
         values.append(v)
 
--- a/FlowCal/io.py
+++ b/FlowCal/io.py
@@ -940,9 +940,9 @@ class FCSFile(object):
     def __hash__(self):
         return hash((self.infile,
                      self.header,
-                     frozenset(iter(self.text.items())),
+                     frozenset(six.iteritems(self.text)),
                      self.data.tobytes(),
-                     frozenset(iter(self.analysis.items()))))
+                     frozenset(six.iteritems(self.analysis))))
 
     def __repr__(self):
         return str(self.infile)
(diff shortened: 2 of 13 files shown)


FlowCal/excel_ui.py
```diff
@@ -133,7 +141,8 @@ def read_table(filename, sheetname, index_col=None):
 
     """
     # Catch sheetname as list or None
-    if sheetname is None or hasattr(sheetname, '__iter__'):
+    if sheetname is None or \
+        (hasattr(sheetname, '__iter__') and not isinstance(sheetname, str)):
```
JS3xton: I'm used to checking `isinstance(other, basestring)`.

This [Stack Overflow post](https://stackoverflow.com/questions/4843173/how-to-check-if-type-of-a-variable-is-string) seems to suggest that checking for a string is different in Python 2.7 versus Python 3.\*. They also mention using the `six` module if you want cross-Python version compatibility? Did you already look into this? I think this ultimately boils down to the current code won't detect `unicode` objects as strings in Python 2.7 (not sure about 3.\*)? (not sure if we care about that)

FlowCal/io.py
```diff
@@ -171,14 +172,18 @@ def read_fcs_text_segment(buf, begin, end, delim=None, supplemental=False):
                              + " TEXT segment")
         else:
             buf.seek(begin)
-            delim = str(buf.read(1))
+            delim = buf.read(1)
+            if sys.version_info.major >= 3:
+                delim = chr(delim[0])
```
JS3xton: How come `delim` needed to be cast to a `chr` explicitly?

test/test_io.py
```diff
@@ -5,7 +5,21 @@
 
 import datetime
 import unittest
-import StringIO
+import sys
+
+# In Python 2.7, StringIO mimics the output of a file stream in 'rb' more (the
```
JS3xton: typo: "more" -> "mode"

FlowCal/excel_ui.py
castillohair: `basestring` doesn't exist in python 3. Conversion tools plainly replace `basestring` for `str`, so that's what I did. I can look into `six`.

JS3xton:
Can you list all of the general substitutions you did? Or point to a good reference that you used?

The couple that I picked out skimming the code changes:
* imports need to be `from . import <module>`
* `dict.iteritems()` -> `iter(dict.items())` - what's the difference between `dict.iteritems()` and `dict.items()` again?
* what's the double `//` division mean?
* `xrange` -> `range` cause `range` in Python 3.0 returns a generator now, right?

castillohair:
- As you said, imports were changed to `from . import <module>`
- Division if floating point by default in python 3. Double division is integer division in both python 2 and 3.
- `xrange` doesn't exist in python 3. `range` returns a generator. So all `xrange` were switched to `range`, and all instances of `range` in which it seemed that a list was expected were changed to `list(range())`.
- `dict.iteritems()` returns an interator in python 2, whereas `dict.items()` returns a list. In python 3 `dict.items()` returns an interatos, and `dict.iteritems()` doesn't exist. These have been changed appropriately. I have seen different methods online on how to manage this, and it seems that people don't necessarily agree on how to do it properly.
- All `buffer.read()` where `buffer` is a file buffer now return either bytes or bytearrays. These cannot be directily compared to strings, so explicit conversion needs to be performed. `test_io` also has to use `BytesIO` instead of `StringIO`.
- There were sevaral instances in which we check for whether something is iterable (like a list), so we checked for `hasattr('__iter__')`. In python 3 strings also contain `__iter__`, so we now have to explicitly check that the element in question is not a string.

FlowCal/io.py
castillohair: `buf.read()` produces a byte (a number), which needs to be casted to a string for downstream use. See my comment in the "Conversation" tab.

FlowCal/excel_ui.py
castillohair: I read more into it, and `six` seems to be a better way to solve this. Alternatively, I could just copy the relevant portion of code:

```python
PY3 = sys.version_info[0] == 3

if PY3:
    string_types = str,
else:
    string_types = basestring
```

I like the second better, since I wouldn't have to worry about a new dependency. Although `six` seems to be included by default in Anaconda, and in a python3.6 virtual machine set up with `pip`.

castillohair: Another option would be to do the following:

```python
if sys.version_info[0]==3:
    basestring = str

if isinstance(s, basestring):
    # do whatever
```

Since `basestring` is not defined in python3, this would work seamlessly.

castillohair:
Regarding `dict`'s iterators, I realized that the current solution is very inefficient in python2. Specifically, `iter(d.items())` produces a list, and then makes an iterator from it. It might be better to just do `d.items()`, which creates a list. The result would be a little more efficient in python2, although not as much as before this pull request. In python3, performance would be unaffected.

Alternatively, we could create the following:
```python
def iteritems(d):
    if sys.version_info.major < 3:
        return d.iteritems()
    else:
        return d.items()

for k, v in iteritems(d):
    # do whatever
```

This would have optimal performance in both python2 and 3.

Finally, it seems that `six` [has a function for this as well](https://pythonhosted.org/six/#six.iteritems).

castillohair:
I found the following about `six`:
- It has been on version 1.10.0 since [2015](https://github.com/benjaminp/six/releases).
- Anaconda has included this version since at least version [4.0.0](https://docs.continuum.io/anaconda/packages/old-pkg-lists/4.0.0/py27) (currently in version 4.4.0).

So it seems to be a library stable enough to not have to worry about it too much. Therefore, I decided to use it.

This has resulted in the following changes:
- All previous calls to `d.iteritems()` now use [`six.iteritems(d)`](https://pythonhosted.org/six/#six.iteritems). This uses `d.iteritems()` in python 2 and `d.items()` in python 3, so it always returns an iterator.
- All previous references to `basestring` and new attempts to check whether an element is a string instance now use [`six.string_types`](https://pythonhosted.org/six/#six.string_types). This uses `basestring` in python 2 and `str` in python 3.
- Calls to `StringIO.StringIO` have been replaced by [`six.BytesIO`](https://pythonhosted.org/six/#six.BytesIO), which returns `StringIO.StringIO` in python 2, and `io.BytesIO` in python 3. This always simulates the output of `file.read()` in `rb` mode.
- The remaining explicit version checks of python now use `six.PY2` (`True` if using python 2, `False` otherwise) and `six.PY3` (`True` if using python 3, `False` otherwise). Currently there is only one of these.

Also, I read more about how strings and unicode are handled in python 3. It seems that there is a parallel between python 2's `str` and python 3's `bytes`, and between python 2's `unicode` and python 3's `str`. See for example the first answer to [this](https://stackoverflow.com/questions/18034272/python-str-vs-unicode-types). I decided to follow this parallel a little closer.

Specifically, I decided that the output of `buf.read()` in `FlowCal.io` should be converted to unicode strings immediately when text is expected (specifically, in functions `read_fcs_header_segment` and `read_fcs_text_segment`. Previously, I was converting to unicode only in `read_fcs_text_segment` and in python 3. This method resulted in some bugs that prevented FlowCal from reading all files in [this repository](https://flowrepository.org/id/FR-FCM-ZZZ4) when using python 3. With the new method, 36/39 files can be read from the repo in both python2 and 3 (same number as with last FlowCal release 1.1.4 in python 2).

Conversion to unicode uses the method `decode(encoding)`. In python 2, this converts a `str` (the output of `read()`) into `unicode`. Similarly, in python 3, `decode()` converts `bytes` into `str`. This results in some degeneracy regarding the unicode encoding (encoding is, roughly, a method to get characters from bytes). The default is [`utf-8`](https://en.wikipedia.org/wiki/UTF-8), which is a multi-byte encoding. This seems to work well with all files tested, except the files from our flow cytometer when we were using the old acquisition card. Specifically, when trying to read file `test/Data001.fcs` with `utf-8`, the following happens: 

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xaa in position 101: invalid start byte
```

Note that 0xAA is 170, a character outside standard ascii. Where is this character coming from? This is the weird � character from the following key/value pair in the head section:
```
CREATOR\CellQuest Pro� 5.1.1
```
(Text extracted using `develop` in python 2. Yes, it looks like a question mark due to not being interpreted correctly)

So an encoding different from `utf-8` has to be chosen. There are [a lot of them](https://docs.python.org/3/library/codecs.html#standard-encodings). I think the most prominent (and the ones mentioned in the introductory unicode documentation in python) are [Code page 437](https://en.wikipedia.org/wiki/Code_page_437) (also known as "extended ascii", although there are many "extended asciis". On windows, you can get a character from this encoding by using ALT+[number]. I had to learn this so I could input things like á and ñ before we got keyboards in spanish) and [ISO 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1), also known as `latin-1`. The latter seems to be more standard. I decided to use this encoding and now nothing fails. The text from the `Data001.fcs` file now reads:

```
CREATOR\CellQuest Proª 5.1.1
```

Which is as good as I'm gonna bother making it.

The most important issue is that, if a file is not in `latin-1`, the file might not get correctly interpreted by FlowCal. This is true regardless of the encoding I chose, and would be true also in the previous python-2-only implementation. To help mitigate this, the `io` module has a module-wide variable called `encoding`. It currently defaults to `'ISO-8859-1'` (same as `latin-1`), but it could be changed by the user in code:

```python
import FlowCal
FlowCal.io.encoding = 'utf-8'
d = FlowCal.io.FCSData('file_in_utf_8.fcs')
```

Finally, the last couple of commits updated the documentation and requirements to reflect that 1) we are now compatible with python 3, and 2) we now require `six` 1.10.0.

JS3xton:
Awesome, this looks well done.

Moving forward, are there any general rules that we as developers now need to adhere to to continue to support both Python 2 and Python 3? (e.g. imports via `from . import <module>`, dictionary iteration via `six.iteritems(dict)`, use of `range` over `xrange` and `list(range())` for an actual list, division via `//` instead of `/`)?

JS3xton (approved):

castillohair:
- Imports inside `__init__.py` should be of the form `from . import <module>`
- Use `//` for integer division. For floating point division, use `/` and convert one of the operands to floating point explicitly.
- Use `list(range())` if you expect to get a list. Never use `xrange`.
- Use `six.iterkeys()`, `six.itervalues()`, and `six.iteritems()` when working with dictionaries.
- Use `six.string_types` to check whether a variable is a string.
- Use `print()` (with parenthesis), which we have been doing anyway.
- Any modifications to the file-reading functions in `io` should be tested in both python2 and 3. From my experience working in this issue, there is no way around spending some time reading about strings, bytearrays, and unicode, and testing in both versions.

That's all I can think of. From now on, I would be ok with doing the main testing/development on python 2, and only testing for functionality on python 3 before releasing a new version.

PR #266, opened 2017-06-19
Use GaussianMixture instead of GMM for sklearn >= 0.18

castillohair:
Solves #230. See issue for testing details.

### DIFF
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -51,9 +51,10 @@ def clustering_gmm(data,
         Number of clusters to find.
     tol : float, optional
         Tolerance for convergence of GMM method. Passed directly to
-        ``scikit-learn``'s GMM.
+        ``scikit-learn``'s GMM object.
     min_covar : float, optional
-        Minimum covariance. Passed directly to ``scikit-learn``'s GMM.
+        Minimum covariance. Passed directly to ``scikit-learn``'s GMM
+        object.
     scale : str, optional
         Rescaling applied to `data` before performing clustering. Can be
         either ``linear`` (no rescaling), ``log``, or ``logicle``.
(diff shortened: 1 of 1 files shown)


FlowCal/mef.py
```diff
@@ -72,14 +80,22 @@ def clustering_gmm(data,
...
 
+    # Initialize min_covar parameter
+    # Parameter is initialized differently depending on scikit's version
+    if min_covar is None:
+        if packaging.version.parse(sklearn.__version__) \
+                >= packaging.version.parse('0.18'):
+            min_covar = 1e-3
+        else:
+            min_covar = 5e-5
+
```
JS3xton, line 100: Why do you need different `min_covar` values for different versions of `sklearn`? Did you come up with the new value for `sklearn >= 0.18` empirically?

JS3xton (approved):
Some general inquiries in the "Conversation" tab and in the code regarding `min_covar`, but otherwise looks good to me.

FlowCal/mef.py
castillohair: There is no `min_covar` argument in [GaussianMixture](http://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html). The closest seems to be `reg_covar`, but it does not have the same effect, based on my testing. With the new method, `min_covar` is only used to initialize the covariance matrix (line 157). Maybe it should have had a better name, but I decided to keep it to not change the function signature.

JS3xton: Ahh. Hmmm, ok. Yeah a better name might help a little there cause that's a little confusing. I don't know if anyone is going to poke around and change the `min_covar` value a lot anyways, though. Up to you. Seems like the documentation should be clarified, though, so that someone can figure out what `min_covar` does for `sklearn >= 0.18` and `sklearn < 0.18`.

castillohair: Just modified the docstring to clarify this.

Issue #277, opened 2017-12-06
FCSData objects are not pickled correctly


SynbioLucas:
## Problem
FCSData objects appear to be incompatible with Python's built-in serialization module, [Pickle](https://docs.python.org/2/library/pickle.html). After serializing and then de-serializing an FCSData instance, the `channels` attribute becomes inaccessible. This may extend to other attributes as well. The `channels` attribute is critical because it is used for slicing and is required to be accessible for FlowCal's built-in plotting methods.

Fixing this issue is important because serialization eliminates the need to go through the extremely slow data extraction procedure each time plots are made.

### Before serialization: works fine
```python
import os, cPickle
import FlowCal as fc

base_dir = os.path.abspath(os.path.join(os.path.abspath('__file__'), '..', '..', 'Flow Data', '20170613'))
fn = os.path.join(base_dir, 'Data001.fcs')

test_fcs = fc.io.FCSData(fn)
print test_fcs.channels
```
#### Output
    ('TIME', 'FSC', 'SSC', 'FL1', 'FL2', 'FL3', 'SSCW', 'SSCA')

### Serialize, deserialize: broken
```python
test_file = os.path.join(os.path.abspath('C:\Users\Lucas\Desktop'), 'test_fcs.p')

with open(test_file, 'w') as outfile:
    cPickle.dump(test_fcs, outfile) # serialization

with open(test_file, 'r') as infile:
    test_fcs = cPickle.load(infile) # de-serialization

print test_fcs.channels
```
#### Output
    ---------------------------------------------------------------------------

    AttributeError                            Traceback (most recent call last)

    <ipython-input-9-34e7fbe9bdc5> in <module>()
    ----> 1 test_fcs.channels
    

    C:\Users\Lucas\Anaconda2\lib\site-packages\FlowCal\io.pyc in channels(self)
       1067 
       1068         """
    -> 1069         return self._channels
       1070 
       1071     def amplification_type(self, channels=None):
    

    AttributeError: 'FCSData' object has no attribute '_channels'

## Notes
- This bug occurs when using both the standard `pickle` module and its C variant, `cPickle`.
- The file read/write mode does not affect this bug (i.e. `'w'` vs `'wb'`).

## Version numbers
- Python: `2.7.13 |Anaconda 4.0.0 (64-bit)| (default, Dec 19 2016, 13:29:36) [MSC v.1500 64 bit (AMD64)]` [from `sys.version`]
- FlowCal: 1.1.3

[...]

PR #278, opened 2017-12-20
Fixes Issue #277. FCSData objects now able to be (de)pickled.

SynbioLucas:
Added code as described in Issue #277 to enable FCSData objects to be pickled and depickled. Wrote a test to verify that this is working properly and that depickled objects contain the correct attributes. All tests pass with no errors.


FlowCal/io.py
```diff
@@ -1760,6 +1760,38 @@ def __array_finalize__(self, obj):
         if hasattr(obj, '_resolution'):
             self._resolution = copy.deepcopy(obj._resolution)
 
+    # override __reduce__ and __setstate__ to fix Issue #277
+    # Adapted from https://stackoverflow.com/a/26599346
```
castillohair: Remove these. I would rather have this explained in a docstring in each function with more detail. See below.

FlowCal/io.py
```diff
@@ -1760,6 +1760,38 @@ def __array_finalize__(self, obj):
...
+    def __reduce__(self):
```
castillohair, line 1779: We need a docstring here explaining what this function does and why it was replaced. It doesn't have to be too detailed, as this function will not be exposed to the user, but it should be enough to be useful for anyone in the future modifying this code. Please refer to https://github.com/numpy/numpy/blob/master/doc/example.py.

FlowCal/io.py
```diff
@@ -1760,6 +1760,38 @@ def __array_finalize__(self, obj):
...
+        fcsdata_attrs = (self._infile, self._text, self._analysis,
+            self._data_type, self._time_step, self._acquisition_start_time,
+            self._acquisition_end_time, self._channels,
+            self._amplification_type, self._detector_voltage,
+            self._amplifier_gain, self._range, self._resolution,)
+        new_state = pickled_state[2] + fcsdata_attrs
+        # Return updated tuple with full set_state tuple
+        return (pickled_state[0], pickled_state[1], new_state)
+
+    def __setstate__(self, state):
```
castillohair, line 1844: We need a docstring here explaining what this function does and why it was replaced. It doesn't have to be too detailed, as this function will not be exposed to the user, but it should be enough to be useful for anyone in the future modifying this code. Please refer to https://github.com/numpy/numpy/blob/master/doc/example.py.

test/test_io.py
```diff
@@ -2681,6 +2682,44 @@ def test_mean_axis(self):
...
+        attrs = set(dir(self.d))
+        loaded_attrs = set(dir(test_fcs))
+        self.assertEqual(attrs, loaded_attrs)
+        # Check contents of attrs affected by pickling
+        pickle_sensitive_attrs = ['_resolution', '_range', '_amplifier_gain',
+            '_detector_voltage', '_amplification_type', '_channels',
+            '_acquisition_end_time', '_acquisition_start_time', '_time_step'
+            '_data_type', '_analysis', '_text', '_infile']
+        for attr in pickle_sensitive_attrs:
+            if hasattr(self.d, attr):
```
castillohair: Why is this necessary? You should know in advance if these attributes are valid or not.

test/test_io.py
```diff
@@ -2681,6 +2682,44 @@ def test_mean_axis(self):
...
+        """
+        channels = self.d.channels # Should be fine, no error
+
+        with open(self.test_file, 'w') as outfile:
+            cPickle.dump(self.d, outfile) # serialization
+
+        with open(self.test_file, 'r') as infile:
+            test_fcs = cPickle.load(infile) # de-serialization
+
+        # Check all attributes/methods are preserved
```
castillohair, line 2716: This section is checking that the same attributes exist, and that some subset of the hidden attributes have the same value. I would recommend that the contents of the array and the actual values of computed attributes are also checked. Even if you know that these worked before this PR, it would be a good idea for the tests to be as comprehensive as possible in order to detect future code changes that break proper pickling.

For the array values, you might want to use [`assert_array_equal`](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.testing.assert_array_equal.html) or [`assert_almost_equal`](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.testing.assert_almost_equal.html).

castillohair:
Thanks for this PR. Indeed all old tests still pass and the Excel UI still works with the example file.

I'm requesting docstrings in the two functions that were added to `FCSData`, and a few additions to the test function.

castillohair (changes requested):
Thanks for this PR. Indeed all old tests still pass and the Excel UI still works with the example file.

I'm requesting docstrings in the two functions that were added to `FCSData`, and a few additions to the test function. After that I think it should be ready to merge, unless @JS3xton has anything else to say.

test/test_io.py
SynbioLucas: These are good ideas - thanks! I believe I've incorporated your suggestions, but please let me know if there are still cases/attributes/data that I'm forgetting!

test/test_io.py
SynbioLucas: Yep, you're right - this is now deleted.

SynbioLucas:
Thanks for your great suggestions. It's obviously been a while since I've worked on this, but would love to still merge if it's possible. I believe I've implemented your requested improvements, but please feel free to let me know if you think of anything else!

Since it's been ages, I also merged in the current `taborlab/master` branch and confirmed that all 365 tests still pass (at least in Python 2.7...).

castillohair:
Made some modifications so that tests pass on python3. They still work on python2.7. @JS3xton if you don't have any other comments, I'd like to merge this now.

castillohair (approved):

test/test_io.py
```diff
@@ -4,8 +4,13 @@
...
 
 import datetime
-import unittest
+import os
 import six
+import unittest
+try:
+   import cPickle as pickle
+except:
+   import pickle
```
JS3xton, line 13: This `except` statement should be as precise as possible. On my computer, you should except a `ModuleNotFoundError`.

SynbioLucas:
While I'm also a fan of specific exception cases, in this particular case,
I think it makes sense to fallback on standard `pickle` no matter what
error was encountered. That said, I don't know of any reason that `cPickle`
would fail to import since it's part of the standard library, but :shrug: I
wrote this way too long ago.

FlowCal/io.py
JS3xton: Consider naming the returned parameters to hint at what they are, e.g.:

        Returns
        -------
        init : callable
            Callable used to recreate the inital instance
        init_args : tuple
            Arguments to be passed to the callable to reconstitute the
            FCSData object
        state : tuple
            Internal state data needed to fully reconstruct the instance. Data
            is passed to self.__setstate__() to reconstruct object during
            deserialization. The order of elements must match that expected
            by __setstate__.

Also, fix the typo with "inital"

FlowCal/io.py
```diff
@@ -1760,6 +1760,97 @@ def __array_finalize__(self, obj):
...
+        fcsdata_attrs = (self._infile, self._text, self._analysis,
+            self._data_type, self._time_step, self._acquisition_start_time,
+            self._acquisition_end_time, self._channels,
+            self._amplification_type, self._detector_voltage,
+            self._amplifier_gain, self._range, self._resolution,)
+        # Append these FCSData-specific state parameters to the existing
+        # NDArray params.
+        new_state = pickled_state[2] + fcsdata_attrs
+
+        return (pickled_state[0], pickled_state[1], new_state)
```
JS3xton: These variable names were confusing to me. References to "state" appear most appropriate for the 3rd tuple element; calling the entire `__reduce__` tuple `pickled_state` therefore sounds imprecise to me.

The [Python 3 docs](https://docs.python.org/3.7/library/pickle.html#object.__reduce__) suggest the output of `__reduce__` is called a "reduce value":
> The __reduce__() method takes no argument and shall return either a string or preferably a tuple (the returned object is often referred to as the “reduce value”).

So consider changing `pickled_state` to something like `super_reduce_value`.

---

Also, the docs suggest that it's possible for `__reduce__` to return up to a 5-tuple, which this code would currently truncate. Can we guarantee that `np.ndarray` will never use those extra tuple elements? Or that it will always already have a "state" element in the 3rd position? I prefer a more precise implementation that is robust to `__reduce__` outputs with between 2 and 5 elements, as the docs allow for.

FlowCal/io.py
```diff
@@ -1760,6 +1760,97 @@ def __array_finalize__(self, obj):
...
+        pickled_state = super(FCSData, self).__reduce__()
+        # Update the __setstate__ state tuple with FCSData attrs
+        fcsdata_attrs = (self._infile, self._text, self._analysis,
+            self._data_type, self._time_step, self._acquisition_start_time,
+            self._acquisition_end_time, self._channels,
+            self._amplification_type, self._detector_voltage,
+            self._amplifier_gain, self._range, self._resolution,)
+        # Append these FCSData-specific state parameters to the existing
+        # NDArray params.
+        new_state = pickled_state[2] + fcsdata_attrs
```
JS3xton: The [docs](https://docs.python.org/3.7/library/pickle.html#object.__reduce__) don't specify how the "state" information is  stored (i.e. in general, we don't know what kind of object `pickled_state[2]` is). For now, the `+` operator appears to append our state information to the parent's state information, but I think it would be more robust to layer our own state information around the parent state information without making any assumptions about the parent state information. E.g. `new_state = (pickled_state[2],fcsdata_attrs)`. Our `__setstate__` function would therefore set our attributes from `new_state[1]` and then pass `new_state[0]` to the `super.__setstate__`.

If we're separating our own state information, you might also consider using a `namedtuple` or `dict` to avoid relying on positional indices to get the attributes unmixed correctly.

FlowCal/io.py
```diff
@@ -1760,6 +1760,97 @@ def __array_finalize__(self, obj):
...
+        self._acquisition_end_time = state[-7]
+        self._acquisition_start_time = state[-8]
+        self._time_step = state[-9]
+        self._data_type = state[-10]
+        self._analysis = state[-11]
+        self._text = state[-12]
+        self._infile = state[-13]
+        # To set remaining state data for parent class, call its
+        # __setstate__ with remaining attrs
+        super(FCSData, self).__setstate__(state[0:-13])
```
JS3xton: As mentioned in `__reduce__`, consider a more robust implementation that packages our state information separate from the parent state information and uses better data structures to avoid relying on position/index to unpack our attributes.

JS3xton: Also, does it matter who does `__setstate__` first? Practically, I think it doesn't matter for this case (?), but philosophically, I kind of feel like the parent class should go first, but I'm not sure.

FlowCal/io.py
castillohair: [Name of return values is optional](https://numpydoc.readthedocs.io/en/latest/format.html), as they're not exposed to the method caller anyway.

castillohair:
@JS3xton, do you wanna make the changes you proposed?

JS3xton:
Alright, I took a stab. It currently passes all unit tests on my machine.

Issue #276, opened 2017-11-01
Long headers in Excel file crash the Excel UI


castillohair:
As title says. In my case, a header such as "520nm Light Intensity (µmol/(m^2*s))" seems to crash the Excel UI. The problem seems to be related to the regex that we use to extract channel names from headers, as they seem to be incurring in [catastrophic backtracking](http://www.regular-expressions.info/catastrophic.html).

Because of my current time availability, I will not be able to provide more details until tomorrow.

castillohair:
(Let's ignore the fact that I said "tomorrow" and it's been almost two months)

The problem is definitely, as I said, a case of [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html). I'll offer a short explanation below, which assumes some familiarity with how the regex engine attempts to match strings. For more information on this, refer to this [regex tutorial](https://www.regular-expressions.info/tutorialcnt.html) or any other regex reference.

Our regex for the "Units" column is currently as follows:
```
^\s*(\S(.*\S)*)\s*Units\s*$
```

With a string such as `520nm Light Intensity (umol/(m^2*s))`, the regex engine has several ways in which it can attempt to match the first portion of the regex pattern (before `\s*Units\s*`). A few expansions of the quantifiers, in the order that the regex engine tries them (as far as I understand) are as follows:

```
"520nm Light Intensity (umol/(m^2*s))"
 S...................................
 S..................................S

 S.................................S.
 S.................................SS

 S................................S..
 S................................S.S
 S................................SS.
 S................................SSS

 S...............................S...
 S...............................S..S
 S...............................S.S.
 S...............................S.SS
 S...............................SS..
 S...............................SS.S
 S...............................SSS.
 S...............................SSSS
```

I will be referring to `"520nm Light Intensity (umol/(m^2*s))"` as the "string", and to each one of my expanded regex patterns as "pattern", starting from the one immediately below the string. The first `S` on each row corresponds to the first `\S` inside the first group (set of parenthesis), before the opening of the second group. Then each sequence of dots corresponds to the `.*` inside the second group, and a `S` at the end corresponds to the `\S` at the end of `(.*\S)`.

A regex engine tries to match as many dots as possible, and when it fails, it will *backtrack* and match the previous number of characters except for the last one. This is what it does in the first pattern in my example. At the end of the first block (second pattern), the engine has successfully matched one repetition of `(.*\S)`, but the whole regex has not matched, so the engine backtracks again.

[...]

Issue #250, opened 2017-04-27
Excel UI should accept channel names (e.g. in column headers) with spaces in them


JS3xton:
The Excel UI should accept channel names (e.g. in column headers) with spaces in them. (currently that seems to fail, in which case the Excel UI ignores [fails to correctly parse] channels with spaces in their labels). Changing the regex in the `excel_ui` module seemed to fix this temporarily. @castillohair proposed `re.compile(r'^\s*(\S.*\S)\s*Units\s*$')`, which seemed to work for my particular use case (although comes with the downside that channel names must have at least 2 characters).

Also might be a useful backup if we could somehow specify the Parameter number instead of the channel name.

JS3xton:
I think `'^\s*(\S(.*\S)*)\s*Units\s*$'` may work. https://docs.python.org/2/howto/regex.html suggests that groups can be nested. `.group(1)` appears to pull out the right channel label (i.e. with spaces). Needs more testing.

JS3xton:
I broke out the parameter number issue into a separate issue (#252) and demoted it to the wishlist.

PR #279, opened 2017-12-29
Regex performance fix

castillohair:
Solves #276 and adds unit tests for Excel UI regexes.

Incorporates test cases from #250 and PR #251 to make sure that these are not reverted after this fix.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -104,8 +104,8 @@ import FlowCal.stats
 import FlowCal.mef
 
 # Regular expressions for headers that specify some fluorescence channel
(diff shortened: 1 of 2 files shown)


JS3xton (approved):
As mentioned in https://github.com/taborlab/FlowCal/issues/276, this is superbly done.

Great idea to formalize the regex interrogations into unit tests.  👍

The one thing I would ask is to show the "before" (6f4993caa390c35486df9d044347f2771dddea7e) and "after" (b67fab2c7e3be6a73439a6d3e806294694468d61) results of the unit tests just to confirm that we see failure before the fix is implemented and passage of all tests after (and because I'm lazy and don't want to install and run them on my own machine. I also think it's just a good idea to have a record of those results anyways). Otherwise looks great.

castillohair:
Test results before regex changes have been documented in #276.

Issue #273, opened 2017-09-22
Make FlowCal More CLI Friendly.


renxida:
# My situation:
This really depends on how each lab organizes their files and use their computers. But I like to do everything command-line, and my lab puts FCS files from one set of experiments inside the same folder and always has flowcal.xlsx as the input file name.

Each time FlowCal is run, I had to use the dialogue that pops out to select the input file, or type in `python -m FlowCal.excel_ui -Hpv -i flowcal.xlsx`.
# Thing I want to change:
1. Make the Run Flowcal shell script pass on arguments with $* to Python, so that i can say `Run\ Flowcal\ (OSX) -i flowcal.xlsx`
2. Add extra script, perhaps named run_flowcal, to quickly run flowcal over `flowcal.xlsx` in the current directory
3. Remove spaces from the names of the Run Flowcal scripts.
4. Add some of the scripts to `.local/bin` when installing on linux.

renxida:
(I'll be submitting a pull request soon. Meanwhile, I would like to know what you think about these changes, especially whether you would object against renaming/adding scripts and adding things to your path when installing flowcal)

castillohair:
Hello,

Thanks for using FlowCal! We hope it's being useful in your research. 

Before I reply to your suggestions point by point, I want to point out the following. We tried to make FlowCal easy to use for as many people as possible, including people who are not familiar with coding or terminal usage. This means that any change that we intend to introduce has to be evaluated in the context of how it is going to affect all of these users. So while I think in general your ideas are good, most of them seem to be specific to your use case, and thus incorporation into FlowCal becomes tricky. In the cases when I think your suggestions have the potential to be incorporated, note that we would require you to provide proof that 1) the old behavior is maintained, and 2) that the new features work as expected.

Now to your suggestions:

>1. Make the Run Flowcal shell script pass on arguments with $* to Python, so that i can say `Run\ Flowcal\ (OSX) -i flowcal.xlsx`

The `Run FlowCal` scripts were not made to be run from the terminal, but just as a convenience for Windows and OSX users to run FlowCal by double clicking. So I would not be completely in favor of a change like this. But if we were to accept it, you would have to show us that the current behavior is preserved. Furthermore, we would like to be consistent and do something similar for the windows version, `Run FlowCal (Windows).bat`. Again, the current behavior should be preserved when calling these without arguments or by double clicking, in all three operating systems. The burden to prove that the scripts perform as expected will be on you before we review a potential pull request.

[...]

PR #280, opened 2017-12-29
Added entrypoints to command line scripts

castillohair:
Solves #273.
- Adds two entrypoints for command line scripts, ``flowcal`` and ``FlowCal``, for the Excel UI CLI that can be used instead of the more verbose ``python -m FlowCal.excel_ui``.
- Reorganized the contents of ``excel_ui.py`` so that processing of command line arguments is made in a separate function that serves as the entry point.
- Updated documentation to reflect new Excel UI CLI.

This is continued from PR #274. A lot of the work was done there by @renxida, but the final code didn't work in my hands. To accelerate iteration, I made a new branch from his commits and committed on top. Original commits and authorship will be preserved after merging of this PR.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -70,6 +70,7 @@ ignored by ``FlowCal``.
 """
 
 import collections
+import sys
 import os
 import os.path
 import packaging
@@ -266,7 +267,7 @@ def process_beads_table(beads_table,
         - Generate forward/side scatter density plots and fluorescence
           histograms, and plots of the clustering and fitting steps of
           standard curve generation, if `plot` = True.
-    
+
     Names of forward/side scatter and fluorescence channels are taken from
     `instruments_table`.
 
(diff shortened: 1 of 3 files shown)


JS3xton (approved):
I mentioned these two issues over in existing discussions in https://github.com/taborlab/FlowCal/pull/274, but I will repeat them briefly here:

1) add brief code comment justifying why we think we can drop the first `sys.argv` element (because `argparse` is doing the same thing if `args=None` for `parser.parse_args()`). Link to source code if possible.
2) remove the `FlowCal` entry point (only permit lowercase `flowcal` entry point) to avoid problems uninstalling the package.

JS3xton:
Also tested 49f28c31d1bab45c478181e133be69ff94510992 on my system (Anaconda on Windows) and was able to run FlowCal from an Anaconda prompt via `flowcal` and `FlowCal` from multiple different directories (although I'm still recommending that we only allow `flowcal`).

I separately verified that the FlowCal outputs were populated in the same location as where the input files are (even if the script is run from a different directory).

castillohair:
Performed both modifications. I found some interesting things while trying to reproduce the issue with the `FlowCal` entrypoint. These have been documented in #274.

Will merge now.

Issue #282, opened 2018-05-02
Pandas argument deprecation on read_excel


castillohair:
Running FlowCal's Excel UI (develop branch dc266f2d520c83068a263cae4240b7837409af4b) on Anaconda 5.1 (python 3.6, pandas 0.22.0) results in the following warning:

```
C:\...\Anaconda3\lib\site-packages\pandas\util\_decorators.py:118: FutureWarning: The `sheetname` keyword is deprecated, use `sheet_name` instead
  return func(*args, **kwargs)
```

It seems to come from file `excel_ui.py`, line 153:

```python
    table = pd.read_excel(filename,
                          sheetname=sheetname,
                          index_col=index_col)
```

A solution to this could check for pandas version and use `sheetname` or `sheet_name`, as appropriate.

castillohair:
This happens with pandas 0.21.0 and later, but not with 0.20.3.

PR #285, opened 2018-06-02
pd.read_excel() is called with different parameters depending on pandas version.

castillohair:
Solves #282.

### DIFF
--- a/FlowCal/excel_ui.py
+++ b/FlowCal/excel_ui.py
@@ -150,9 +150,17 @@ def read_table(filename, sheetname, index_col=None):
         raise TypeError("sheetname should specify a single sheet")
 
     # Load excel table using pandas
-    table = pd.read_excel(filename,
-                          sheetname=sheetname,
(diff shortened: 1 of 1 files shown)


castillohair:
This is a pretty straightforward version checking. If there are no comments until Tuesday morning, I'll merge.

JS3xton:
Is there any reason you went with explicit version checking over a `try`-`except` approach? Explicit version checking feels more precise, but less Pythonic. I'm OK with whatever you wanna do here, though.

castillohair:
- This is not the first time we do things differently depending on some module's version. These cases have been handled via explicit version checking. Example: https://github.com/taborlab/FlowCal/pull/266/files. So consistency is one reason why I'm using this method.
- `try-except` blocks should be used when there is more uncertainty about the failure mode, I think, since such method can catch failures other than what we intend. Here, we know exactly what statement can fail and when (which depends on `pandas`' version). So explicit checking makes more sense to me.

Issue #283, opened 2018-05-02
Pandas deprecation of set_value


castillohair:
Running FlowCal's Excel UI (develop branch dc266f2d520c83068a263cae4240b7837409af4b) on Anaconda 5.1 (python 3.6, pandas 0.22.0) results in the following warning:

```
...\FlowCal\excel_ui.py:982: FutureWarning: set_value is deprecated and will be removed in a future release. Please use .at[] or .iat[] accessors instead
```

The warning is issued several times during a single run. `DataFrame.set_value()` was the recommended way to modify cells in a dataframe, and it is what we have used in `excel_ui`.

`DataFrame.set_value()` is used 18 times in `excel_ui.py`, so checking for the appropriate version of pandas and using `DataFrame.set_value()` or `DataFrame.at[]` in each one of these is a little inconvenient. An alternative approach would be to define a module-level function `excel_ui.set_value()` that uses `DataFrame.set_value()` or `DataFrame.at[]` as appropriate, so that the rest of `excel_ui` uses `excel_ui.set_value()` to modify dataframe's cells.

JS3xton:
I'm not in love with a module-level function, but I also don't have a great alternative. I agree that polluting the code with a bunch of `pandas` version checks probably makes the code base less readable, which is undesirable. It might be prudent to be more specific with the function name, though. I was originally thinking `excel_ui.set_dataframe_value()`, but this deprecation appears to affect [Series, Panel, and SparseSeries](https://medium.com/@agrimabahl/deprecations-in-pandas-v-0-22-and-how-to-fix-them-3bc8935c9612) too. We've introduced the concept of a "table" via `read_table()`, `process_beads_table()`, `process_samples_table()`, `generate_histograms_table()`, `generate_about_table()`, so maybe `excel_ui.set_table_value()`? It's also unclear to me whether this function should be publicly exposed or hidden.

castillohair:
I like `excel_ui.set_table_value()`. I think I'll move on to implement a solution with this approach.

castillohair:
This happens with pandas 0.21.0 and later, but not with 0.20.3.

PR #286, opened 2018-06-02
excel_ui uses df.set_value() or df.at[] depending on pandas' version.

castillohair:
Solves #283.

I ended up checking for pandas' version on every case. The fact that a bunch of assignments were clustered together made it a little easier.


castillohair:
This is a pretty straightforward version checking. If there are no comments until Tuesday morning, I'll merge.

JS3xton:
Is there any reason you went with explicit version checking over a `try`-`except` approach? Explicit version checking feels more precise, but less Pythonic. I'm OK with whatever you wanna do here, though.

castillohair:
Same as #285.

Issue #284, opened 2018-05-30
Deprecation of matplotlib._cntr (Density Gate)


castillohair:
`gate.density2d()` uses the `matplotlib._cntr` module to generate the contour output. However, this module has been deprecated in matplotlib 2.2.0 (see https://github.com/matplotlib/matplotlib/issues/10709/).

There's a module intended as a replacement build by matplotlib people, available [here](https://github.com/matplotlib/legacycontour). It doesn't give the impression that it is fully working right now. I was able to find a different module that seems to perform the same function [here](https://ccarocean.github.io/python-contours/). 

If we can't find a direct replacement, our alternatives are:
- Code the contour methods ourselves without external libraries
- Do not support matplotlib 2.2.0 until this is resolved.

JS3xton:
I skimmed the underlying C implementation behind `matplotlib._cntr` and was convinced that it would be nontrivial to reproduce the algorithm ourselves in Python.

From https://github.com/matplotlib/matplotlib/issues/10709, a [relevant comment](https://github.com/matplotlib/matplotlib/issues/10709#issuecomment-383130722) suggests using `_contour` in place of `_cntr` (even though it's still a private function). I haven't tested that, though. That comment also points to an older relevant issue (https://github.com/matplotlib/matplotlib/issues/367), which documents requests to separate contour calculation from plotting (exactly what we want). One of the `matplotlib` developers points to `scikit-image` for that functionality, though, and the thread hasn't really gone anywhere since then.

After a cursory investigation, it looks like `scikit-image` might do what we want in a more publicly sanctioned way:

http://scikit-image.org/docs/dev/auto_examples/edges/plot_contours.html

It would require another dependency, but it looks like `scikit-image` is a dependable package, so that gets my vote at the moment.

castillohair:
In principle, it seems silly to me to use a whole new package for a small part of a single function in FlowCal (remember that no other gating function requires something like this). The more packages we have, the more work we have to put in when their features are deprecated or names are changed (which happens more often that we would like). On the other hand, we do use `scikit-learn` for a single machine learning method, so there's precedent of using a whole module for a single purpose, although I would argue that clustering is a more complicated task that determining a contour.

[...]

PR #288, opened 2018-07-10
Generation of contours for plotting is carried out by scikit-image.

castillohair:
Solves #284.

### DIFF
--- a/FlowCal/gate.py
+++ b/FlowCal/gate.py
@@ -506,15 +506,23 @@ def density2d(data,
     gated_data = data[mask]
 
     if full_output:
-        # Find contours with scikit-image
-        # x and y coordinates of each input value are its row and column
-        # indices. Output values are interpolated from the grid of row and
-        # column indices, and should be converted to associated x and y values.
+        # Use scikit-image to find the contour of the gated region
+        #
+        # To find the contour of the gated region, values in the 2D probability
+        # mass function ``D`` are used to trace contours at the level of the
+        # probability associated with the last accepted bin, ``vD[sidx[Nidx]]``.
+
+        # find_contours() specifies contours as collections of row and column
+        # indices into the density matrix. The row or column index may be
+        # interpolated (i.e. non-integer) for greater precision.
+        contours_ij = skimage.measure.find_contours(D, vD[sidx[Nidx]])
+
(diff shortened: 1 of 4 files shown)


JS3xton:
How did you arrive at the `>=0.10.0` version requirement?

castillohair:
I could not install any previous version via ``pip``. That said, those versions may work, but if they can't be installed via our installation mechanisms then I see no reason to support them.

FlowCal/gate.py
```diff
@@ -506,32 +506,24 @@ def density2d(data,
...
+        # Get contours in row/column space
+        contours_ij = skimage.measure.find_contours(D, vD[sidx[Nidx]])
+        # Transform into x/y coordinates
+        contours = [np.array([np.interp(contour_ij[:,0],
+                                        np.arange(len(xc)),
+                                        xc),
+                              np.interp(contour_ij[:,1],
+                                        np.arange(len(yc)),
+                                        yc)]).T
+                    for contour_ij in contours_ij]
```
JS3xton, line 532: It took me a while to figure out what was going on here. More verbose/explicit comments might help. I took a stab:

```python
    if full_output:
        # Use scikit-image to find the contour of the gated region
        #
        # To find the contour of the gated region, values in the 2D density
        # array are used to trace contours at the level of the probability 
        # associated with the last accepted bin.

        # find_contours() specifies contours as collections of row and column
        # indices into the density matrix. The row or column index may be
        # interpolated (i.e. non-integer) for greater precision.
        contours_ij = skimage.measure.find_contours(D, vD[sidx[Nidx]])

        # Map contours from indices into density matrix to histogram x and y
        # coordinate spaces (assume values in the density matrix are associated
        # with histogram bin centers).
        xc = (xe[:-1] + xe[1:]) / 2.0   # x-axis bin centers
        yc = (ye[:-1] + ye[1:]) / 2.0   # y-axis bin centers

        contours = [np.array([np.interp(contour_ij[:,0],
                                        np.arange(len(xc)),
                                        xc),
                              np.interp(contour_ij[:,1],
                                        np.arange(len(yc)),
                                        yc)]).T
                    for contour_ij in contours_ij]
```

JS3xton (approved):

PR #290, opened 2018-07-10
set_xscale() and set_yscale() are called with arguments only if scale is logicle

castillohair:
Fixes #289.

Issue #287, opened 2018-06-03
`normed` argument in `plt.hist()` has been changed to `density`.


castillohair:
Starting `matplotlib` `2.2.0`, the following warning is thrown when calling `plot.hist1d()`:

```
/usr/local/lib/python3.6/dist-packages/matplotlib/axes/_axes.py:6462: UserWarning: The 'normed' kwarg is deprecated, and has been replaced by the 'density' kwarg.
```

(`2.1` does not present this warning).

The problem seems to arise from `plt.hist()`, as the following line throws the same warning:

```python
plt.hist(d, normed=True)
```

castillohair:
7f27b8cd93ad66bfdef57458dc85d12f2e3f5a82 fixes this issue. The example files can be run without the error message above under matplotlib 2.2.2.

The fix involves using the `density` argument if matplotlib's version is greater than 2.2 or equal, and normed otherwise. Note that the signature of `FlowCal.plot.hist1d()` does not change, and the parameter `normed_area` is mapped to either `normed` or `density`.

Furthermore, I checked that `normed_area` worked as before. The following was run with file 
[Data002.fcs.zip](https://github.com/taborlab/FlowCal/files/2178852/Data002.fcs.zip).

```python
import FlowCal

d = FlowCal.io.FCSData('Data002.fcs')
d = FlowCal.transform.to_rfi(d)

FlowCal.plot.hist1d(d,
                    channel='FL1',
                    xscale='log',
                    normed_area=True,
                    savefig='hist_normed_area.png')
```

With matplotlib 2.1.2 and commit 7f27b8cd93ad66bfdef57458dc85d12f2e3f5a82, where `normed` should be used when calling `plt.hist()`:

![hist_normed_area](https://user-images.githubusercontent.com/1529481/42489878-3c24fd6a-83d3-11e8-99c3-0114cbfd1fb5.png)

With matplotlib 2.2.2, and without the proposed fix:

![hist_normed_area](https://user-images.githubusercontent.com/1529481/42489893-551f15ee-83d3-11e8-999e-d0eda0ec1479.png)

The error message specified above appears here, as expected.

With matplotlib 2.2.2 and commit 7f27b8cd93ad66bfdef57458dc85d12f2e3f5a82:

![hist_normed_area](https://user-images.githubusercontent.com/1529481/42489922-6d5b849e-83d3-11e8-845e-656dddc9a9a0.png)

No warning or error message appears here.

[...]

PR #291, opened 2018-07-10
Use `normed` or `density` for `plt.hist()` depending on matplotlib's version

castillohair:
Solves #287.

Note that branch had to be based off ``log-scale-fix`` instead of ``develop`` because it is impossible to test without the changes therein. If #290 is merged into develop, the section "files changed" will look as intended. I recommend doing #290 first.

### DIFF
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -46,6 +46,7 @@ Functions in this module are divided in two categories:
 
 """
 
+import packaging
 import numpy as np
 import scipy.ndimage.filters
 import matplotlib
@@ -795,8 +796,10 @@ def hist1d(data_list,
         Label to use on the x axis. If None, attempts to extract channel
         name from last data object.
     ylabel : str, optional
-        Label to use on the y axis. If None and ``normed==True``, use
-        'Probability'. If None and `normed==False``, use 'Counts'.
+        Label to use on the y axis. If None and ``normed_area==True``, use
+        'Probability'. If None, ``normed_area==False``, and
+        ``normed_height==True``, use 'Counts (normalized)'. If None,
+        ``normed_area==False``, and ``normed_height==False``, use 'Counts'.
     xlim : tuple, optional
         Limits for the x axis. If not specified and `bins` exists, use
         the lowest and highest values of `bins`.
@@ -891,23 +894,43 @@ def hist1d(data_list,
             weights = None
 
         # Actually plot
-        if bins is not None:
-            n, edges, patches = plt.hist(y,
-                                         bins,
-                                         weights=weights,
-                                         normed=normed_area,
-                                         histtype=histtype,
-                                         edgecolor=edgecolor[i],
-                                         facecolor=facecolor[i],
-                                         **kwargs)
+        if packaging.version.parse(matplotlib.__version__) \
+                >= packaging.version.parse('2.2'):
+            if bins is not None:
+                n, edges, patches = plt.hist(y,
+                                             bins,
+                                             weights=weights,
+                                             density=normed_area,
+                                             histtype=histtype,
+                                             edgecolor=edgecolor[i],
+                                             facecolor=facecolor[i],
+                                             **kwargs)
+            else:
+                n, edges, patches = plt.hist(y,
+                                             weights=weights,
+                                             density=normed_area,
+                                             histtype=histtype,
+                                             edgecolor=edgecolor[i],
+                                             facecolor=facecolor[i],
+                                             **kwargs)
         else:
-            n, edges, patches = plt.hist(y,
-                                         weights=weights,
-                                         normed=normed_area,
-                                         histtype=histtype,
-                                         edgecolor=edgecolor[i],
-                                         facecolor=facecolor[i],
-                                         **kwargs)
+            if bins is not None:
+                n, edges, patches = plt.hist(y,
+                                             bins,
+                                             weights=weights,
+                                             normed=normed_area,
+                                             histtype=histtype,
+                                             edgecolor=edgecolor[i],
+                                             facecolor=facecolor[i],
+                                             **kwargs)
+            else:
+                n, edges, patches = plt.hist(y,
+                                             weights=weights,
+                                             normed=normed_area,
+                                             histtype=histtype,
+                                             edgecolor=edgecolor[i],
+                                             facecolor=facecolor[i],
+                                             **kwargs)
 
     # Set scale of x axis
     plt.gca().set_xscale(xscale, data=data_list, channel=channel)
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -56,6 +56,7 @@ import matplotlib.ticker
 import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 from matplotlib.font_manager import FontProperties
+import warnings
 
 # Use default colors from palettable if available
 try:
@@ -837,7 +838,20 @@ def hist1d(data_list,
     passed to ``plt.hist``. Additional keyword arguments provided to
     `hist1d` are passed directly to ``plt.hist``.
 
+    If `normed_area` is set to True, `hist1d` calls ``plt.hist`` with
+    ``density`` (or ``normed``, if matplotlib's version is older than
+    2.2.0) set to True. There is a bug in matplotlib 2.1.0 that
+    produces an incorrect plot in these conditions. We do not recommend
+    using matplotlib 2.1.0 if `normed_area` is expected to be used.
+
     """
+    # Using `normed_area` with matplotlib 2.1.0 causes an incorrect plot to be
+    # produced. Raise warning in these conditions.
+    if normed_area and packaging.version.parse(matplotlib.__version__) \
+            == packaging.version.parse('2.1.0'):
+        warnings.warn("bug in matplotlib 2.1.0 will result in an incorrect plot"
+            " when normed_area is set to True")
+
     # Convert to list if necessary
     if not isinstance(data_list, list):
         data_list = [data_list]
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -910,7 +910,10 @@ def hist1d(data_list,
                                          **kwargs)
 
     # Set scale of x axis
-    plt.gca().set_xscale(xscale, data=data_list, channel=channel)
+    if xscale=='logicle':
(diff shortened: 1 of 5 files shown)


FlowCal/plot.py
```diff
@@ -891,26 +894,49 @@ def hist1d(data_list,
...
-            n, edges, patches = plt.hist(y,
-                                         bins,
-                                         weights=weights,
-                                         normed=normed_area,
-                                         histtype=histtype,
-                                         edgecolor=edgecolor[i],
-                                         facecolor=facecolor[i],
-                                         **kwargs)
+        if packaging.version.parse(matplotlib.__version__) \
+                >= packaging.version.parse('2.2'):
```
JS3xton, line 912: Is there a reason you chose to use the `density` keyword for `mpl 2.2` and above? Looks like it was introduced in 2.1: https://matplotlib.org/users/prev_whats_new/whats_new_2.1.0.html#density-kwarg-to-hist , so you could maybe more appropriately use it there too. This is a pretty narrow use case, though, and not terribly important.

JS3xton:
Adding version checks bloats the code and doesn't scale well with time. If we forewent the version check and only used the `density` keyword, though, we would only be able to support back to `mpl 2.1`? Is that right? Whereas we currently support back to `mpl 1.3.1`? And we think it's worth the addition of version checks in the code to support older `mpl` versions? Given that the `density` keyword is still pretty new, I think I buy that. (We'd only be able to support back to Anaconda version 5.0.1, and I'm still routinely using version 4.4.0).

JS3xton:
I'm still trying to unravel the behavior of `FlowCal.plot.hist1d()` across all the supported versions of underlying packages.

- It looks like the `density` and `normed` keywords do the same thing back to [`mpl 2.0.0`'s `hist`](https://matplotlib.org/2.0.0/api/_as_gen/matplotlib.axes.Axes.hist.html) (even though [`density` was only introduced in `mpl 2.1.0`](https://matplotlib.org/users/prev_whats_new/whats_new_2.1.0.html#density-kwarg-to-hist)), and that behavior is to calculate a valid probability density that integrates to 1 (even with variable bin widths). So our `normed_area` keyword always does the same thing with these `mpl` versions.
- In `mpl 1.5.0` (the major `mpl` release prior to `mpl 2.0.0`, which we still claim to support), `mpl` inherits `numpy` behavior? (see Notes section of documentation from [`mpl 2.0.0`'s `hist`](https://matplotlib.org/2.0.0/api/_as_gen/matplotlib.axes.Axes.hist.html) ) And up until `numpy 1.5`, the `normed` keyword was "buggy" for variable width bins? So if you try really hard, you might get different `FlowCal.plot.hist1d()` behavior with `mpl <= 1.5` and `numpy <= 1.5` with variable bin widths? Looks like `numpy <= 1.5` is pretty old (we require `numpy>=1.8.2`, [Anaconda v1.0 starts with ~`numpy 1.7`](https://docs.anaconda.com/anaconda/packages/old-pkg-lists/1.0))

So my understanding is that we should have consistent behavior from `FlowCal.plot.hist1d()` (specifically with variable width bins) across all supported `numpy` and `mpl` versions, but only because we don't support older `numpy` versions.

---
Also, what happens when `normed_area` and `normed_height` are both `True`? Our docs say `normed_height` is ignored, but that's not done explicitly in the code (both flags perform their duties and pass data structures to `plt.hist`). The documentation for `plt.hist` suggests that specifying `weights` and `density` is a valid input combination; is it a special degenerate case when `weights` are all the same (crafted to normalize the bin heights) that doing the `density` calculation effectively ignores `weights`? (I haven't tested this or done the math myself). Favoring more explicit, readable code, and given the unstable nature of `plt.hist`, we may want to make that behavior more explicit. Maybe changing:

```python
        # Decide whether to normalize
        if normed_height:
            weights = np.ones_like(y)/float(len(y))
        else:
            weights = None
```

to

```python
        # Decide whether to normalize
        if normed_height and not normed_area:
            weights = np.ones_like(y)/float(len(y))
        else:
            weights = None
```
(haven't tested this, though)

FlowCal/plot.py
castillohair: Two reasons:
- I saw no warning when running with matplotlib 2.1. If it doesn't complain, I'm not gonna fix it.
- More importantly, `plt.hist()` in matplotlib `2.1.0` does not interpret the `normed` or `density` arguments properly. More details here: https://github.com/matplotlib/matplotlib/issues/9557/ In the case of FlowCal, `plot.hist1d()` with `normed_area=True` does not produce the same plot with `matplotlib 2.1.0` as it does with any other version that I've tested, including `2.0`, `2.1.2`, and `2.2.x`. Again, this is due to an intrinsic problem with `plt.hist()`, therefore I think it's not something we can address without calculating the histogram ourselves. And I don't think we should do that for a matplotlib bug that has already been solved. Either way, what we do for matplotlib 2.1.0 doesn't matter because the result is wrong anyway.

castillohair:
> Adding version checks bloats the code and doesn't scale well with time. If we forewent the version check and only used the density keyword, though, we would only be able to support back to mpl 2.1? Is that right? Whereas we currently support back to mpl 1.3.1? And we think it's worth the addition of version checks in the code to support older mpl versions? Given that the density keyword is still pretty new, I think I buy that. (We'd only be able to support back to Anaconda version 5.0.1, and I'm still routinely using version 4.4.0).

And as I said above, matplotlib 2.1 does not calculate normalized histograms properly. So if we eliminate version checks, we will effectively support matplotlib versions from 2.1.2 only. I think this is too restrictive. Also, we've been dealing with arising incompatibilities via version checks so far, so doing that here is consistent. If we want to remove this, it would be more appropriate to do it everywhere, and this may be outside the scope of this issue. Again, I don't think it's worth it.

> So my understanding is that we should have consistent behavior from FlowCal.plot.hist1d() (specifically with variable width bins) across all supported numpy and mpl versions, but only because we don't support older numpy versions.

Yeah that's what I think too. I can test later with the absolute oldest requirements and confirm.

> is it a special degenerate case when weights are all the same (crafted to normalize the bin heights) that doing the density calculation effectively ignores weights?

That's what makes sense to me mathematically. I could test later.

> Favoring more explicit, readable code, and given the unstable nature of plt.hist, we may want to make that behavior more explicit.

I agree that this is a good idea, but probably outside of the scope of this PR. This one should only deal with argument deprecation.

castillohair:
- With the oldest possible requirements (`matplotlib 1.3.1`, `numpy 1.8.2`, etc) the plot that results with `normed_area` is the same as with every other version I've tested, when using current branch `hist-normed` (see below, compare to plots in #287)

hist_normed_area


- To test how both `normed_height` and `normed_area` affect the output, I wrote the following script:

```python3
import FlowCal

d = FlowCal.io.FCSData('Data002.fcs')
d = FlowCal.transform.to_rfi(d)

FlowCal.plot.hist1d(d,
                    channel='FL1',
                    xscale='log',
                    savefig='hist_not_normed.png')

FlowCal.plot.hist1d(d,
                    channel='FL1',
                    xscale='log',
                    normed_height=True,
                    savefig='hist_normed_height.png')

FlowCal.plot.hist1d(d,
                    channel='FL1',
                    xscale='log',
                    normed_area=True,
                    savefig='hist_normed_area.png')

FlowCal.plot.hist1d(d,
                    channel='FL1',
                    xscale='log',
                    normed_height=True,
                    normed_area=True,
                    savefig='hist_normed_both.png')
```

The results were as follows:

Not normed:

hist_not_normed

With `hist_normed_height`:

hist_normed_height

With `hist_normed_area`:

hist_normed_area

With both:

hist_normed_both

We can see that the output with both of these artuments is as if only `normed_area` had been set.

castillohair:
If there are no more comments by the end of the weekend, I'll merge.

JS3xton:
@castillohair and I spoke about this in person yesterday. @castillohair reiterated that `mpl 2.1.0` causes `FlowCal.plot.hist1d()` to yield incorrect results. We don't know about `mpl 2.1.1`; @castillohair said he could test that. @castillohair said `mpl 2.1.2` was verified as working correctly, though.

I briefly went through other plotting code and convinced myself that only `FlowCal.plot.hist1d()` is vulnerable to the `mpl 2.1.0` histogram error. `FlowCal.plot.density2d()`, for example, uses `np.histogram2d()` (not matplotlib) to calculate the histogram, and then it visualizes it via `plt.scatter()` or `plt.pcolormesh()`. We are unaware of issues with any of these functions as we are using them.

@castillohair and I agreed that, in lieu of re-implementing histogram plotting localy if the matplotlib version is detected to be 2.1.0, the responsible thing to do is document the matplotlib error in the `FlowCal.plot.hist1d()` description and raise a Warning if FlowCal detects that the user may be generated an inconsistently illustrated FlowCal histogram (i.e. if the matplotlib version is detected to be 2.1.0 and the user has indicated that the histogram should be normalized).

castillohair:
Plot produced with matplotlib 2.1.1 is also ~~in~~correct.

(Edited because I messed up)

castillohair:
I was able to find the histogram function code for different versions or matplotlib and find the offending code. I have a better idea of why the `density` argument broke when it did, and when it was fixed. For the record, `hist()` is coded as `Axes.hist()` in `lib/matplotlib/axes/_axes.py`.

Here's a diff of the relevant section of `pyplot.hist()`, left: code in `matplotlib 2.0.2`, right: `matplotlib 2.1.0`. 

2 0 2 to 2 1 0

On the left, `hist_kwargs`  is passed to `np.histogram` in line 323. Before that, in line 314 they mention that they're not passing the `normed` argument to `np.histogram` (and note that line 313 is commented), and taking care of normalization in lines 327-329. On the right, `density` is passed to `np.histogram` (lines 285 and 294), but normalization is carried anyway in lines 298-300. This is consistent with [this comment](https://github.com/matplotlib/matplotlib/issues/9557#issuecomment-339081722).

Below, there's a diff between versions 2.1.0 and 2.1.1:

2 1 0 to 2 1 1

You can see that the offending code that performs normalization for the second time has been removed. As I indicated above, `FlowCal`'s `hist1d` with `normed_area=True` are produced correctly with `matplotlib 2.1.1`.

This further confirms that the only matplotlib version that handles this incorrectly is 2.1.0.

castillohair:
Added warning code about matplotlib 2.1.0. I'll merge tomorrow night if there are no more comments.

JS3xton (approved):

PR #305, opened 2019-07-03
Updated code in python API tutorial.

### LINKED ISSUE #304: Python API Tutorial needs updating
There are some minor things like ```print x``` statements that don't work with Python 3. There's also the following statement caught by Moshe:

```python
>>> # Obtain transformation function
>>> # The following MEFL values were provided by the beads' manufacturer
>>> mefl_values = [0, 646, 1704, 4827, 15991, 47609, 135896, 273006]
>>> to_mef = FlowCal.mef.get_transform_fxn(b_g,
...                                        mef_values=mefl_values,
...                                        mef_channels='FL1',
...                                        plot=True)
>>> plt.show()
```

Which produces the following error:

```
.../FlowCal\mef.py in get_transform_fxn(data_beads, mef_values, mef_channels, clustering_fxn, clustering_params, clustering_channels, statistic_fxn, statistic_params, selection_fxn, selection_params, fitting_fxn, fitting_params, verbose, plot, plot_dir, plot_filename, full_output)
   1066         # Get selected rfi and mef values
   1067         selected_rfi = stats_values[selected_mask]
-> 1068         selected_mef = mef_values_channel[selected_mask]
   1069
   1070         # Accumulate results

TypeError: only integer scalar arrays can be converted to a scalar index
```

Which could be easily fixed by changing `mefl_values = [0, 646, 1704, 4827, 15991, 47609, 135896, 273006]` to `mefl_values = np.array([0, 646, 1704, 4827, 15991, 47609, 135896, 273006])`, which is what the current example scripts do.

Overall, the whole tutorial needs to be double checked against the most recent package versions.

castillohair:
Fixes #304 .

All the python tutorial was rerun with the latest package versions.

Since this is an easy fix, I'll give 24h for validation before merging.

### DIFF
--- a/doc/python_tutorial/mef.rst
+++ b/doc/python_tutorial/mef.rst
@@ -7,8 +7,9 @@ To start, navigate to the ``examples`` directory included with FlowCal, and open
 
 >>> import FlowCal
 
-Also, import ``pyplot`` from ``matplotlib``
+Also, import ``numpy`` and ``pyplot`` from ``matplotlib``
 
+>>> import numpy as np
 >>> import matplotlib.pyplot as plt
 
 Working with Calibration Beads
@@ -47,7 +48,7 @@ Generating a transformation function from calibration beads data is a complicate
 
 >>> # Obtain transformation function
 >>> # The following MEFL values were provided by the beads' manufacturer
->>> mefl_values = [0, 646, 1704, 4827, 15991, 47609, 135896, 273006]
+>>> mefl_values = np.array([0, 646, 1704, 4827, 15991, 47609, 135896, 273006])
 >>> to_mef = FlowCal.mef.get_transform_fxn(b_g, 
 ...                                        mef_values=mefl_values,
 ...                                        mef_channels='FL1',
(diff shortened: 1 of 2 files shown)


JS3xton (commented):
Why can't `FlowCal.mef.get_transform_fxn()` take `mef_values=list()`? Seems like `FlowCal.mef.get_transform_fxn()` should be made more robust. The example should then be updated to reflect the suggested usage (not sure if that's `list` or `np.array()`).

castillohair:
> Why can't `FlowCal.mef.get_transform_fxn()` take `mef_values=list()`? Seems like `FlowCal.mef.get_transform_fxn()` should be made more robust.

See #308. We can keep the tutorial fix (this PR) and fixing `mef_values` as separate. 

> The example should then be updated to reflect the suggested usage (not sure if that's `list` or `np.array()`).

The function's docstring says that `mef_values` should be a "sequence of sequences", so either should be fine. However, the example scripts use `np.array`, so I would prefer to change this piece of code to `np.array` as well to make them uniform.

JS3xton (approved):

Issue #306, opened 2019-07-03
Move documentation to readthedocs


castillohair:
https://readthedocs.org is a free service that hosts documentation of open source packages. Hosting our documentation website there seems to have several advantages:
- The source of our Sphinx-based documentation can be automatically read, compiled, and published whenever we want to. This can even be done automatically every time we push into master.
- We won't have to rely on the awkward method based on the `gh-pages` branch, which is completely different from the repo itself but needs to be kept as a branch for our current website to work.
- Documentation from several package versions can be maintained simultaneously.

I have implemented a pilot version at https://flowcal.readthedocs.io. The url should be final but the repo is currently pointing at my own fork. This is because I needed to add a few files to `master`.

Given that we can have documentation of many FlowCal versions available, @JS3xton and I talked about retroactively generating documentation for old FlowCal versions. However, given that a few text files need to be retroactively added to make this work, we would have to modify the repo's history, which will be a messy process. Thus, we think the best course of action is to keep FlowCal documentation published starting with the current version.

The current plan of action is:
- Make any modifications that are necessary on the repo files in a new feature branch, as usual. These will be tested in the master branch of my own fork of FlowCal. These will also include changing every reference to https://taborlab.github.io/FlowCal/ into https://flowcal.readthedocs.io.
- Submit a PR once I verify that everything works.
- Once the PR (and #305) are approved, I will publish a new version (with version number increment) immediately.
- I will redirect the FlowCal project on `readthedocs` to this repo.
- I will delete the gh-pages branch.
- I will notify Felix so that the lab website is updated with the new documentation url.

castillohair:
Moving to readthedocs completed.

PR #307, opened 2019-07-03
Modifications to work with Readthedocs

castillohair:
Solves #306.

Summary of changes:
- Added necessary `.readthedocs.yml` file.
- Added `requirements.txt` file in the `doc` folder, to be used by readthedocs when compiling the documentation.
- Modified one docstring in `excel_ui.py` that was incompatible with the current version of `numpydoc`, a package used to generate the documentation.
- Changed the documentation url in the README file.

Documentation can be seen at https://flowcal.readthedocs.io. However it currently points to my own fork. This will be changed as soon as this PR is accepted and a new FlowCal version is released.

Since this is a pretty minor change, I will merge in 24h if there are no comments.


JS3xton (changes requested):
`readthedocs` appears to [serve ads on your site](https://docs.readthedocs.io/en/latest/advertising/ethical-advertising.html)... That's a real turn off for me. You realize and are OK with that??

castillohair:
I'm fine with this for the following reasons:
- The single ad they show is limited to one specific portion of the screen, below the navigation panel. It doesn't get in the way of the content or, worse, pretends to be content (*cough* twitter *cough*).
- They claim that they don't store personal information or build a personal profile for ad targeting, which are my main problems with modern online ads.
- They seem to be one of the largest repositories of open source documentation, and a significant fraction of the python packages I've used use it ([example](https://openpyxl.readthedocs.io/en/stable/), [another example](https://six.readthedocs.io/index.html#), [yet another example](https://readthedocs.org/projects/pip/)). I would expect this to put pressure on them to not scale up the obnoxiousness of their ads. But if this proves to be wrong over time, I'm happy to find somewhere else.

We could opt out of commercial advertising and [only show ads that promote other open source projects or conferences](https://docs.readthedocs.io/en/stable/advertising/ethical-advertising.html#community-ads). I'm fine with this, if you think this will ease your concerns a little bit. We could also pay $5 a month and not show ads.

Finally, take into account that this reduces the work that has to be done every time there's a new release. As the person having to do said work, showing one ad sounds like a more than acceptable tradeoff.

castillohair:
We got authorization to use readthedocs, therefore I'm merging this PR.

PR #311, opened 2019-08-03
Fixed compatibility issue with matplotlib 3.1.x.

castillohair:
Fixes #310.

PR #315, opened 2019-08-03
Locator tick fix in matplotlib >= 3.1.0

### LINKED ISSUE #314: Messed up tick values in logicle plots
With matplotlib version 3.1.0 and above, there are many weird overlapping tick values on the lower side of each axis when using the logicle scale. Example:

clustering_B0001

castillohair:
Fixes #314 

Tested with matplotlib 3.0.0 and 2.0.0 as well.

#311 should be merged before this.


castillohair:
If there are no further comments, I will merge during the weekend.

JS3xton (approved):

Issue #317, opened 2019-10-14
MacOS annoyances


castillohair:
1. The new version of MacOS has changed its default shell to [zsh](https://support.apple.com/en-us/HT208050). Attempting to run the "Run FlowCal (OSX)" script now results in the following error:

```
-bash: /.../FlowCal/Run FlowCal (OSX): /bin/sh^M: bad interpreter: No such file or directory
```

It seems that this is a result of windows-style line endings that we've been using, which bash was able to parse correctly but zsh can't. I propose to explicitly change the 2 macOS scripts line endings to unix-style using the [.gitattributes file ](https://stackoverflow.com/questions/10357436/how-to-make-git-not-change-line-endings-for-one-particular-file). Just changing the files' endings will not work, because git seems to change line endings back to windows-style when committing.

2. macOS makes these `.DS_Store` files that it puts on every folder. Every time I run `git status` I get the following:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.DS_Store
	examples/.DS_Store
```

I would like to add `.DS_Store` to `.gitignore` so this is not an issue anymore.

3. The name of the operating system is nowadays known as macOS, not OSX. I would like to change the installation script names to reflect this.

PR #316, opened 2019-10-11
Fixed file select box hanging after file selection on OSX. 

felixekn:
Added a tkinter root window for the file select box to be placed into. This prevents the file select box from hanging after a file is selected. Tested on MacOSX 10.14.6 in python 3.7.4. Fixes #299

PR #318, opened 2019-10-14
MacOS issues

castillohair:
Solves #317. Necessary to test #316 in macOS Catalina 10.15.


castillohair:
Will give 24h for any comments, since this is a pretty minor change.

JS3xton:
Changing the script names seems like an API change warranting a major version bump to me. Have you starting thinking about the next release?

castillohair:
I don't think this qualifies as a change in the API (Application *Programming* Interface). Plus, the name of the file can be easily changed by users if the filename is really important to them.

PR #323, opened 2020-03-09
feat: add channel_labels

abawchen:
Add `channel_labels` and pass unittest.

### DIFF
--- a/test/test_io.py
+++ b/test/test_io.py
@@ -24,11 +24,13 @@ Files to test:
     - Data003.fcs: FCS 3.0 from FlowJo Collectors Edition 7.5 /
                     BD FACScan Flow Cytometer
     - Data004.fcs: FCS 3.0 including floating-point data
+    - Data005.fcs: FCS 3.0 including $PxS value
 """
 filenames = ['test/Data001.fcs',
             'test/Data002.fcs',
             'test/Data003.fcs',
             'test/Data004.fcs',
+            'test/Data005.fcs',
             ]
 
 class TestFCSDataLoading(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


abawchen:
@castillohair @JS3xton : Please help review this one, thanks.

JS3xton (changes requested):
Thanks for the contribution. I found some small things, but otherwise looks good to me.

1) The `channel_labels` method should be described in the `FCSData` docstring.
2) It would be nice to explicitly unit test `channel_labels` like other similar methods (e.g. the [TestFCSAttributesDetectorVoltage](https://github.com/taborlab/FlowCal/blob/9cd83ef7fadf235e548bc6c1bb5ed22d4dde8416/test/test_io.py#L1584) class in `/test/test_io.py`).

JS3xton:
3. Add `channel_labels` method to [`TestFCSDataPickle`](https://github.com/taborlab/FlowCal/blob/9cd83ef7fadf235e548bc6c1bb5ed22d4dde8416/test/test_io.py#L2690) in `/test/test_io.py`.

abawchen:
@JS3xton : Thanks for feedback, I will update accordingly.

abawchen:
@JS3xton : I have a quick question, because all of the current testing fcs files don't have label value, i.e. they don't have `$PnS` key, and will return `[None, None, ..., None]` of all channels -- should I modify the current testing fcs file or create another one?

JS3xton:
> @JS3xton : I have a quick question, because all of the current testing fcs files don't have label value, i.e. they don't have `$PnS` key, and will return `[None, None, ..., None]` of all channels -- should I modify the current testing fcs file or create another one?

@abawchen Hmm, good point. I wouldn't modify the existing files. If you have a new file you can contribute, that seems fine to me. Otherwise, you might be able to find one in this repository: https://flowrepository.org/id/FR-FCM-ZZZ4.

castillohair:
@JS3xton , I added the description of `channel_labels` into the `FCSData` docstring. I also added it to the `TestFCSDataPickle` test.

@abawchen I think we would still prefer if you could incorporate an FCS file that actually contains `$PnS` keys and make unit tests for it. Code looks good, but we still have no "experimental" evidence that it does what it's supposed to do.

abawchen:
@JS3xton @castillohair :
Hi, thanks your feedback, although it's been a while 😅 
I have added `Data005.fcs` and add corresponding test case as you suggested, please help check it.

btw, I have rebase the updated develop branch.

test/test_io.py
```diff
@@ -2301,6 +2303,146 @@ def test_nondefault_nbins_many_5(self):
                                 scale='linear'),
             [bins1, np.arange(262145) - 0.5, bins2])
 
+class TestFCSAttributesChannelLabels(unittest.TestCase):
+    """
+    Test correct extraction, functioning, and slicing of channel label.
+
+    We have previously looked at the contents of the $PnS attribute for
+    the test files and identified the correct label(marker):
+        - Data001.fcs: [None, None, None, None, None, None]
```
castillohair, line 2312: The last element of this list should be `'Time (204.80 sec.)'`. This is what you have later in the `test_attribute()` function.

castillohair:
In my opinion, the proposed unit tests validate that the feature works as expected. I have one observation in the test class docstring though. @JS3xton see what you think.

JS3xton (approved):

JS3xton:
Looks good to me.

(The new data file is kinda large, though... ~16MB)

castillohair:
@abawchen thanks for the contribution! We will release it with the new FlowCal version soon.

Issue #181, opened 2016-01-19
viridis as default colormap for density plots?


JS3xton:
The colormaps here: https://bids.github.io/colormap/ seem well thought out. Looks like they already exist in matplotlib 1.5 and will be the default in matplotlib 2.0, so using them might eliminate a dependency on `palettable`.

Looks like we're currently using `palettable.colorbrewer.diverging.Spectral_8_r.mpl_colormap`, which may achieve the same things that the ones above do, I don't know. I'm also not sure how complicated it is to support the colormaps mentioned above during this transition period. May be better to wait until they're the matplotlib default, and then update FlowCal to embrace them?

castillohair:
I think the first step would be to generate a density plot and a couple of clustering plots, to see if we like the result and we want to go further with this. I personally really like the current clustering plots. The different colors indicate different discrete groups, but there is also an order associated with them. So I wouldn't care about perceptual uniformity here. But I can see the benefit of changing the 2D density plots.

There is also the issue of the standard curve plots, which would revert to the plain blue, red, and green used by `matplotlib` by default. This means that we would still need `palettable` for this plot. They are changing the default styles in `matplotlib 2.0` (http://matplotlib.org/style_changes.html), but I can't find specifics. You might wanna spend some time in the github repo (https://github.com/matplotlib/matplotlib) and see if you can find something. I'd like to have this information and see a sample plot before proceeding any further.

castillohair:
On the standard curve plots, it turns out matplotlib includes styles besides the classic one. This includes some styles from `seaborn` and `ggplot`. Some of the `seaborn` styles are similar to what is currently done for the standard curves. See http://matplotlib.org/users/style_sheets.html.

castillohair:
Now that some time has passed and I have some experience with viridis, I have to say that I'm warming up to this idea. Defaults in matplotlib > 2.0.0 are pretty nice and could justify ditching palettable. If anyone wants to do this, I'd be interested. I probably won't do it, though.

PR #328, opened 2020-05-06
Remove palettable

JS3xton:
Remove dependency on `palettable` and replace it with similar alternatives from `matplotlib`. Closes https://github.com/taborlab/FlowCal/issues/181.

Comparison of plots produced by `>python -m FlowCal.excel_ui -p -i ./examples/experiment.xlsx`:
| Before pull request (9cd83ef7fadf235e548bc6c1bb5ed22d4dde8416) | After pull request (e44e9c8ef59b04f5a45646c0e6b897426224d241)|
|:----------------------:|:--------------------:|
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

I.e. blue now looks purple, but otherwise it's similar.

### DIFF
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -24,7 +24,7 @@ import FlowCal.plot
 import FlowCal.transform
 import FlowCal.stats
 
-standard_curve_colors = ['b', 'g', 'r']
+standard_curve_colors = ['tab:blue', 'tab:green', 'tab:orange']
 
 def clustering_gmm(data,
                    n_clusters,
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -24,14 +24,7 @@ import FlowCal.plot
 import FlowCal.transform
 import FlowCal.stats
 
-# Use default colors from palettable if available
-try:
-    import palettable
-except ImportError as e:
-    standard_curve_colors = ['b', 'g', 'r']
-else:
-    standard_curve_colors = \
-        palettable.colorbrewer.qualitative.Paired_12.mpl_colors[1::2]
+standard_curve_colors = ['b', 'g', 'r']
 
 def clustering_gmm(data,
                    n_clusters,
--- a/FlowCal/mef.py
+++ b/FlowCal/mef.py
@@ -24,7 +24,7 @@ import FlowCal.plot
 import FlowCal.transform
 import FlowCal.stats
 
-standard_curve_colors = ['tab:blue', 'tab:green', 'tab:orange']
+standard_curve_colors = ['tab:blue', 'tab:green', 'tab:red']
 
 def clustering_gmm(data,
                    n_clusters,
--- a/FlowCal/plot.py
+++ b/FlowCal/plot.py
@@ -58,7 +58,7 @@ from mpl_toolkits.mplot3d import Axes3D
 from matplotlib.font_manager import FontProperties
 import warnings
 
-cmap_default = plt.get_cmap(matplotlib.rcParams['image.cmap'])
+cmap_default = plt.get_cmap('Spectral_r')
 
 savefig_dpi = 250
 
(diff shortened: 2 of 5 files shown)


castillohair:
Besides the individual comments, I like the results in most cases. Going all the way to purple allows better differentiation between the first few subpopulations.

The exception is 1D histograms. This might be subjective, but I don't like that the first plot is purple. Would it be better to have 1D histograms follow matplotlib's default color cycle? Meaning first one would be blue, second orange, third green, etc.

JS3xton:
Sorry, I don't completely understand your comment @castillohair.

I understand you like purple for some plots but not others? Specifically, you don't like purple for the 1D histograms (e.g. bottom plot of last image? and maybe 2nd and 3rd plots of second image?). The third image is 1D histograms but they correspond with the subpopulations in the first image and color coordination is helpful. What do you propose there?

Recall these colors all come from a [module-level `cmap_default` variable](https://github.com/taborlab/FlowCal/blob/e44e9c8ef59b04f5a45646c0e6b897426224d241/FlowCal/plot.py#L61). If you want to use different color maps for different sets of plots, that would probably have to change. Or are you saying _all_ colormaps should be the `matplotlib` default?

I personally don't have strong feelings either way regarding the blue -> purple change. I have a weak preference for preserving previous behavior (i.e. using `Spectral_r`, which resembles the previous colormap, over the `matplotlib` default). I don't love the purple, but I figured I'd get used to it.

We could also probably make a slightly modified `Spectral_r` colormap that doesn't go all the way to purple, which would even more closely resemble the old colormap. I prefer the simplicity of embracing the standard `Spectral_r` map, though, over making a custom solution.

castillohair:
I'm talking specifically about 1D histograms that result when calling `density_and_hist()`, i.e. 2nd and 5th rows in your table. These are currently generated from `cmap_default` as follows:

```python
colors = [cmap_default(i) for i in np.linspace(0, 1, n_colors)]
```

But there's no reason these should come out of a colormap because they don't appear in the same axis. So I propose these to be taken from the current color cycle, which would normally result in blue, orange, green, red, etc. It seems that this could be done as follows:

```python
colors = ['C{}'.format(i) for i in range(n_colors)]
```

See [this](https://matplotlib.org/3.1.1/api/colors_api.html) and [this](http://blog.rtwilson.com/easily-specifying-colours-from-the-default-colour-cycle-in-matplotlib/). Although I don't know if these would roll over if you have more plots than the color cycle, which is what happens when you call `plt.plot()` with too many lines. Also, which is the oldest matplotlib version that supports this.

Everything else looks great to me.

JS3xton:
New commit (c02b5d0898adc03e04e24e890c5f873c714a2d9f) should achieve what you've asked for.

I used 
```python
default_property_cycler = plt.rcParams['axes.prop_cycle']()
colors = [next(default_property_cycler)['color'] for i in range(n_colors)]
```

instead of

```python
colors = ['C{}'.format(i) for i in range(n_colors)]
```

which should wrap around. E.g.


```python
n_colors = 15
default_property_cycler = plt.rcParams['axes.prop_cycle']()
colors = [next(default_property_cycler)['color'] for i in range(n_colors)]

# colors = ['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd', '#8c564b', '#e377c2', '#7f7f7f', '#bcbd22', '#17becf', '#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd']
```

New plots 2 and 5:

JS3xton:
The images in the documentation should also probably be updated once we have the behavior (colormaps) we want.

castillohair:
Changed standard curve colors for modern equivalents.

Current develop (9cd83ef7fadf235e548bc6c1bb5ed22d4dde8416):
std_crv_FL1_B0001

After @JS3xton 's last commit (c02b5d0898adc03e04e24e890c5f873c714a2d9f):
std_crv_FL1_B0001 copy

After my update (833a099daaa0f2c9325dea524e830a7368398e68):
std_crv_FL1_B0001 SMC


Also confirmed that everything works with our oldest matplotlib supported version, 2.0.0.

Will merge now.

